PR #28, opened 2026-07-25
refactor(implement): trim the description; the body already meets the new rubric

ericrisco:
Fourth per-skill pass. Reviewed `implement` against the rubric from #23 — **most of the body passes as written**, which is worth recording as much as a big diff would be.

Two things I expected to cut and deliberately did not:

- **`Anti-patterns → STOP` stays.** The rubric *requires* an anti-patterns table (dimension 3). It is a different animal from the rationalization tables removed from `init` and `harness`: those restated rules stated directly above them, this one names concrete failure modes with their fixes.
- **The accompaniment-dial table stays.** Unlike the one removed from `harness`, it is not a re-teaching of the dial — it says what to *show at each implementation checkpoint*, which is specific to this phase and documented nowhere else.

**Description 743 → 509 chars.** Trigger list out; the progress-ledger behaviour in. Surviving compaction is the non-obvious property that distinguishes this phase, and it was missing from a description that had room for ten trigger phrases — a good illustration of what the old rubric was buying us.

Verification: `eval-lint` PASS (6 should_trigger / 5 should_not_trigger / 1 capability), `npm run validate` OK, 194 tests green.

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -2280,7 +2280,7 @@
     },
     {
       "id": "implement",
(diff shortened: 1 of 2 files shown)


ericrisco:

PR #31, opened 2026-07-25
refactor(ab-testing): discriminative description, drop restated summary sections

ericrisco:
Context-engineering pass on `skills/ab-testing/SKILL.md` against the updated `scripts/skill-rubric.md`. **No technical substance changed** — every formula, default, library call, number and recommendation is byte-identical to before.

## Metrics

| | Before | After |
|---|---|---|
| `description` chars | 746 | **350** |
| SKILL.md bytes | 12012 | **9831** |
| SKILL.md lines | 191 | **159** |
| `scripts/eval-lint.sh` | PASS | **PASS** (should_trigger=7, should_not_trigger=5, capability=1) |

> Note: the commit body quotes 9789 bytes / 158 lines — those were pre-final-edit figures and the amended commit could not be pushed (force-push is blocked by policy). The correct figures are the ones in the table above.

## Description

Before (746 chars) carried a nine-phrase `Triggers: '...'` keyword list. After (350 chars):

> Use when designing or analyzing a controlled experiment — falsifiable hypothesis, sample size from an MDE, reading significance/CI/power, CUPED, or rescuing tests that won't go significant. NOT recurring metric tracking (that is `analytics`), NOT north-star/KPI trees (that is `kpi-framework`), NOT projecting metrics forward (that is `forecasting`).

All three `NOT` siblings verified to exist under `skills/`, and all three are genuine near-misses in `evals/cases.yaml`, so each boundary clause earns its characters.

## Removed

- **`## When to use`** (5 bullets) — restated the description almost verbatim; its reference pointers duplicated links already sitting at point of use in Steps 2/3/5.
- **The trailing `## Decision table`** (9 rows) — a summary, not a branch point. Rows 1–3 duplicated the Step 4 metric-type table exactly; the rest restated Step 3 (sequential vs fixed horizon) and Step 5 (CUPED yes/no) prose that already explains *why*, which teaches judgment better than a bare row.
- **The standalone `## References` index** — both files are now linked inline where the reader needs them, with the same contents summary folded into the Step 2 and Step 3 pointers. Nothing became unlinked.
- One line of stance prose ("Do the math first…") that the anti-patterns table and Step 3 already enforce concretely.

## Deliberately kept

- **The pre-test checklist.** It overlaps Steps 1–3 by design: it is the emittable pre-flight gate, the steps are the reasoning behind it.
- **The Step 4 metric-type table** — the flow genuinely branches there.
- **The full anti-patterns table** (rubric requires one; all 9 rows are distinct failure modes, none a restatement of a step).
- **Every `Why:` clause next to the rule it governs** (peeking inflates Type-I, ~1/MDE² scaling, 3 primaries ≈ 14% FPR). Load-bearing, non-obvious, and already adjacent to their step rather than in a separate rule bank — so no rule-bank surgery was needed; this skill never had "Iron rules" or a "Rationalizations — STOP" table.
- **The `## Checkable artifact` note** — `scripts/verify.sh` is undiscoverable otherwise.

## Verification

- `bash scripts/eval-lint.sh | grep '^ab-testing'` → `PASS (should_trigger=7, should_not_trigger=5, capability=1)`
- `description` is 350 chars (≤350 aim, ≤1024 hard limit) and parses as single-line quoted YAML
- Both `references/pitfalls.md` and `references/sample-size-and-cuped.md` are linked from the body
- All code fences language-tagged (`text`, `python`, `python`, `text`)
- `manifest.json` untouched (regenerated by the orchestrator); `npm test` / `npm run validate` not run (no node_modules in this worktree)

### DIFF
--- a/skills/ab-testing/SKILL.md
+++ b/skills/ab-testing/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: ab-testing
-description: "Use when designing or analyzing a controlled experiment — writing a falsifiable hypothesis, sizing a test from a minimum detectable effect, reading significance/confidence intervals/power, or rescuing a test that 'won't go significant'. Triggers: 'how many users do I need to detect a 2% lift', 'is this result significant', 'control 540/10000 vs variant 590/10000', 'why is my A/B test inconclusive', 'we keep peeking and stopping early', 'set up CUPED to speed up the test', 'sample ratio mismatch', 'cuántos usuarios necesito para el test', 'el experiment no surt significatiu'. NOT recurring metric tracking (that is analytics), NOT defining north-star/KPI trees (that is kpi-framework), NOT projecting a metric forward (that is forecasting)."
+description: "Use when designing or analyzing a controlled experiment — falsifiable hypothesis, sample size from an MDE, reading significance/CI/power, CUPED, or rescuing tests that won't go significant. NOT recurring metric tracking (that is `analytics`), NOT north-star/KPI trees (that is `kpi-framework`), NOT projecting metrics forward (that is `forecasting`)."
 tags: [ab-testing, experimentation, statistics, cuped, sample-size, hypothesis-testing]
 recommends: [analytics, kpi-framework, forecasting, data-cleaning, python, reporting]
 origin: risco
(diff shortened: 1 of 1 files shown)


ericrisco:

PR #32, opened 2026-07-25
refactor(analyze): cut the description 940→434 chars and drop redundant rule banks

ericrisco:
Context-engineering pass on `skills/analyze/SKILL.md` against the updated `scripts/skill-rubric.md`. Phase semantics preserved exactly — `analyze` is still the report-only pre-implementation gate, still runs the same six analyses, still writes the same report.

## Numbers

| | Before | After |
|---|---|---|
| `description` | 940 chars | **434 chars** |
| SKILL.md | 13,958 bytes / 141 lines | **10,377 bytes / 106 lines** |
| eval-lint | — | **PASS** (`should_trigger=7, should_not_trigger=6, capability=1`) |

## Description

Deleted the 11-phrase `Triggers: '...'` keyword list — the model matches by meaning, and this is a `core`-profile skill, so that list was paid for in context on every turn it is installed, invoked or not. Rebuilt as what/when plus a four-way discriminative boundary against real siblings: `tasks` (the phase before), `implement` (the phase after), `verify` (the post-code gate people confuse with this pre-code one) and `clarify` (which *resolves* the ambiguity analyze only *reports*).

## Removed

- **"When to use / when NOT to use"** (10 lines) — the description now carries the boundary; the missing-artifact case was already handled in Inputs.
- **"Red flags — stop and report"** (6 lines) — three of four bullets restated Inputs or the report-only rule. The one genuinely new bullet (findings pouring in across all six checks → re-run `clarify`/`plan` first) was folded into Output §4, next to the routing it governs.
- **Model-routing mechanics paragraph** — `sdd` owns that method; replaced by one line naming this phase's tier (`heavy`) and pointing at `../sdd/references/model-routing.md`.
- **Three of eight anti-pattern rows** that duplicated other rows or restated "run all six analyses".
- **The standalone "The one rule that defines this skill" heading** — the rule is kept, compressed into one paragraph next to the chain position, saying *why* it is absolute instead of shouting NON-NEGOTIABLE.

## Deliberately kept

- Report-only discipline; the six analyses including the carrier-completeness / isolated-implementer check; the requirement-coverage map; the severity scale; the report + OKF v0.1 output contract; the next-in-chain hand-off. Wording unchanged where it was already tight — this is a context pass, not a content rewrite.
- **Anti-patterns table** (rubric requires one), reframed from rationalization→reality to anti-pattern→why it breaks the gate.
- **The accompaniment dial**, compressed to a pointer plus an L0..L3 table: that L-level mapping is *this* phase's report-rendering contract (asserted by the capability eval), not a restatement of what `init` owns.

## Verification

- `bash scripts/eval-lint.sh | grep '^analyze'` → `PASS`
- description 434 chars (≤1024 hard limit), parses as single-line quoted YAML
- all four sibling links resolve (`../sdd/SKILL.md`, `../sdd/references/model-routing.md`, `../plan/references/plan-template.md`, `../tasks/SKILL.md`)
- no `references/` dir in this skill, so nothing is left unlinked
- `manifest.json` untouched (derived; orchestrator regenerates)

Note for the orchestrator: `analyze` has **no result-envelope block** — it never had one, unlike `specify`/`tasks`/`implement`/`verify`/`ship`. Adding one would be a content change, so it was left out of this pass; worth a follow-up if the chain expects envelopes from every phase.

### DIFF
--- a/skills/analyze/SKILL.md
+++ b/skills/analyze/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: analyze
-description: "Use when the spec, plan and task list are written and you want a last consistency check BEFORE any code is touched — the pre-implementation GATE of the rsc-sdd chain. It cross-reads constitution vs spec vs plan vs tasks and reports gaps, contradictions, duplication, ambiguity and scope drift, mapping every artifact to the requirements it claims to satisfy. Report-only: it never edits artifacts and never writes code; the user decides each resolution. Triggers: 'analyze', 'analiza la spec antes de implementar', 'cross-check constitution spec plan tasks', 'consistency gate', 'are spec and plan aligned', 'did we miss any requirement', 'scope drift check', 'gap analysis before coding', 'is the plan complete', 'sanity-check before implement', 'coverage of requirements'. NOT clarifying a single spec (that is `clarify`), NOT writing the plan or tasks, NOT running tests/lint (that is `verify`), NOT root-causing a bug (that is `debug`)."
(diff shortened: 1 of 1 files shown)


ericrisco:

PR #33, opened 2026-07-25
refactor(api-design): cut the trigger-phrase list from the description and move reference pointers to the step that needs them

ericrisco:
Context-engineering pass on `skills/api-design/` against the updated `scripts/skill-rubric.md`. **No technical substance changed** — every recommendation, status code, RFC, header, version and command is byte-identical to before.

## Numbers

| Metric | Before | After | Δ |
|---|---|---|---|
| `description` chars | 914 | **387** | −527 (−58%) |
| SKILL.md bytes | 13132 | **12416** | −716 (−5.5%) |
| SKILL.md lines | 189 | **182** | −7 |
| `eval-lint.sh` | PASS | **PASS** | 7 should_trigger / 5 should_not_trigger / 1 capability |

## What was removed

- **The `Triggers: '...'` phrase list** (11 quoted phrases, EN + ES). The description sits in context on *every* turn the skill is installed, invoked or not, and the model matches by meaning — enumerating surface forms was pure per-turn cost.
- **The `NOT receiving inbound webhooks` clause.** `webhooks` is the least confusable of the four boundaries and is still delegated by name in the Handoff section.
- **The trailing `## References` index.** All four reference files were already pointed at inline from the section that needs them, so the index was a second copy of the same pointers. The content signals it carried that were *not* already inline (content negotiation, rate-limit headers, HATEOAS note, the three versioning mechanisms, what `scripts/verify.sh` checks) were folded into the inline pointers rather than dropped, and all four are now real markdown links.

## What was deliberately kept

This skill was **already written to the new shape**. There were no `Iron rules (non-negotiable)` / `Rationalizations — STOP` banks to delete, no re-teaching of another skill's material, and no accompaniment-dial duplication. So the honest diff is small:

- **The Anti-patterns table** — the rubric requires one.
- **The REST-vs-GraphQL and pagination decision tables** — the flow genuinely branches there.
- **The Bad/Good URL table alongside the four modeling rules.** They look redundant but are not: the rules carry PUT-vs-PATCH semantics and the query-string grammar example (`?status=open&sort=-created_at&fields=id,title`), neither of which the table states.
- **The `Your one job` framework-neutrality paragraph** — it is the scope boundary that stops this skill drifting into `fastapi`/`nestjs` territory.

## Note on the 350-char aim

The description landed at **387**, over the ≤350 aim by 37 and far under the 1024 hard limit. Held there on purpose: the three retained boundaries (`fastapi`/`nestjs`/`go`/`nodejs`, `secure-coding`, `api-connector-builder`) are exactly the near-misses the eval `should_not_trigger` set names. Trimming to 350 meant dropping one of them, which costs discriminative power — the dimension the rubric actually scores.

## Verification

- `bash scripts/eval-lint.sh | grep '^api-design'` → `api-design PASS (should_trigger=7, should_not_trigger=5, capability=1)`
- `description` = 387 chars, parses as single-line quoted YAML, third person, `Use when` lead, explicit `NOT … (that is <sibling>)` boundaries.
- All 4 files under `references/` are linked from the body.
- All 8 `../<sibling>/SKILL.md` links resolve; all 8 `recommends` ids are real skills.
- All 5 opening code fences language-tagged.

### DIFF
--- a/skills/api-design/SKILL.md
+++ b/skills/api-design/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: api-design
-description: "Use when designing the contract of a network API before/independent of implementation — modeling resources and URLs, choosing REST vs GraphQL, picking a versioning strategy, defining one consistent error envelope, and settling pagination/filtering/idempotency conventions. Triggers: 'design the REST API for X', 'what should the endpoints be', 'how do I version this without breaking clients', 'should this be REST or GraphQL', 'cursor vs offset pagination', 'standard error format for my API', 'RFC 9457 problem details', 'idempotency key on POST', '200 with error body', 'diseña la API REST de esto', 'cómo verziono la API sin romper clientes'. NOT implementing endpoints in a framework (that is fastapi/nestjs/go/nodejs), NOT security/auth hardening (that is secure-coding), NOT receiving inbound webhooks (that is webhooks), NOT building an outbound client to a third-party API (that is api-connector-builder)."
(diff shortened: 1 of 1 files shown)


ericrisco:

PR #35, opened 2026-07-25
refactor(plan): discriminative description, progressive-disclosure body

ericrisco:
Context-engineering pass on `skills/plan/SKILL.md` against the updated `scripts/skill-rubric.md`. SDD phase semantics are unchanged — this is presentation, not content.

## Numbers

| | before | after |
| --- | --- | --- |
| `description` | 881 chars | **418 chars** |
| `SKILL.md` | 14489 bytes / 221 lines | **9471 bytes / 148 lines** |

## Description

Deleted the 9-phrase `Triggers: '…'` keyword list (pure per-turn cost — the model matches by meaning). Kept an explicit `NOT <x> (that is \`sibling\`)` boundary against all four neighbours — `specify` and `clarify` before, `tasks` and `implement` after — so the phase is choosable from the description alone.

> Use when an approved, clarified spec must become a technical blueprint before any code — the SDD phase between clarify and tasks: architecture, contracts, data flow, testing strategy, sequencing and risks, at structure altitude, never framework syntax. NOT the what/why spec (that is `specify`), NOT the ambiguity sweep (that is `clarify`), NOT the task breakdown (that is `tasks`), NOT the code (that is `implement`).

## Removed

- **§1–§7 per-section prose** that restated, sentence for sentence, the guidance already in `references/plan-template.md` — which is linked and loaded exactly when a plan is being written. Only the three sections that fail quietly (architecture, testing strategy, risks) stay inline.
- The **duplicated architecture diagram** (template §2 has it) and the prose duplicate of the `reserve` contract example.
- **Re-teaching owned elsewhere**: the model-routing protocol mechanics (`sdd` owns them → pointer to `../sdd/references/model-routing.md`) and what the accompaniment dial is (`init` owns it → pointer).
- The **"Rationalization" framing** of the anti-patterns table, reworded to anti-pattern → why it fails / fix.

## Deliberately kept

- The **anti-patterns table** (rubric-required), all 8 rows — each routes or prevents a real failure.
- The **entry gate** with both STOP branches (no spec → `specify`, unclarified spec → `clarify`). It is the phase's routing contract; the reason is now stated ("every phase downstream inherits the guess") rather than shouted.
- **"Decide structure, not syntax"** plus the language-neutral contract example — this phase's entire altitude.
- The section table, the artifact/slug rule ("one plan per spec, same name"), the plan-specific dial table, the always-propose-isolation block, and the chain position.

One row **added**: `§0 Global Constraints`. It is already in `references/plan-template.md` and `tasks`/`implement`/`analyze` all depend on the plan carrying it, but the body's section table never mentioned it.

## Verification

- `bash scripts/eval-lint.sh | grep '^plan '` → `plan PASS (should_trigger=7, should_not_trigger=6, capability=1)`
- `description` 418 ≤ 1024, parses as single-line quoted YAML
- `references/plan-template.md` is linked from the body (the only reference file)
- Every `../<sibling>/SKILL.md` link resolves: clarify, constitution, fastapi, flutter, go, init, nextjs, postgresdb, specify, tasks, worktrees
- `manifest.json` untouched (regenerated by the orchestrator)

### DIFF
--- a/skills/plan/SKILL.md
+++ b/skills/plan/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: plan
-description: "Use when turning an approved, clarified spec into a technical implementation plan — the SDD phase between clarify and tasks. Covers architecture and component boundaries, the public interfaces/contracts, data model and data flow, the testing strategy, sequencing, and the risk register. Triggers: 'plan this feature', 'how should we build this', 'design the architecture', 'write the implementation plan', 'plan.md', 'what's the data flow', 'how do we test this', 'what are the risks', 'we have the spec, now what'. Reads the spec under 02-DOCS/wiki/sdd/specs/ and the constitution; writes 02-DOCS/wiki/sdd/plans/<slug>.md; defers concrete stack mechanics (framework APIs, ORM, test tooling) to the relevant stack skill. NOT requirements capture (that's specify), NOT ambiguity removal (that's clarify), NOT the task breakdown (that's tasks). Honors the harness accompaniment dial."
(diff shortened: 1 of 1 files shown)


ericrisco:

PR #37, opened 2026-07-25
refactor(api-connector-builder): migrate to the Claude-5-era authoring rubric

ericrisco:
## Metrics

| | Before | After |
|---|---|---|
| `description` chars | 891 | **349** (target ≤350, hard limit 1024) |
| Body bytes | 16032 | **14505** |
| Body lines | 285 | **268** (ceiling 400) |

## What went

- **The `Triggers:` phrase list** — seven quoted prompts across EN/ES/CA. The model matches by meaning; a keyword list is pure per-turn cost on every install, invoked or not. That single deletion is most of the 542 chars.
- **The "Operating posture" bullet bank.** Four of five rules were already stated where they govern: env-only secrets, per-request timeouts and "never log the token" are rows in the anti-patterns table with identical consequences, and "read the vendor docs first" is what Step 0 is. The unique one — *one connector = one vendor* — moved into the intro beside the deliverable it constrains, per "a rule in context is followed, a rule in an appendix is skimmed".
- **The "this skill is outbound, webhooks is the inverse" paragraph** — a third restatement of a boundary the description and the anti-patterns already carry.
- **The trailing reference index.** Both files were already linked inline at point of use, so the index was a second copy of the same two links; the one detail only it carried (dedup-on-overlap) was folded into the inline pagination mention. Every file under `references/` remains linked from the body.

## What stayed, and why

- **The anti-patterns table** — required by the rubric, and a different animal from a rule bank: it names concrete failure modes (re-POST without an idempotency key, `for page in range(N)`, token in query string) rather than repeating rules stated above it.
- **Every decision table** — auth-flow selection, retry-vs-don't by status, rate-limit header precedence, pagination style. The flow genuinely branches at each.
- **Both bad/good code pairs** (retry loop, rate-limit wait) and the two-language worked connector — judgment taught by demonstration, not padding.
- **All sourced claims verbatim** — OAuth 2.1 deltas, RFC 9700 refresh rotation, AWS Builders' Library on jitter and idempotency, tenacity 9.1.4 / undici, with their access dates.

Substance is unchanged: no recommendation, version, figure, command or claim was altered. Only `skills/api-connector-builder/SKILL.md` is touched — `manifest.json` left for the orchestrator to regenerate.


ericrisco:

PR #40, opened 2026-07-25
refactor(ai-media): migrate to the Claude-5-era authoring rubric

ericrisco:
## Numbers

| | Before | After |
|---|---|---|
| `description` chars | 762 | **344** |
| SKILL.md bytes | 14679 | **13331** |
| SKILL.md lines | 197 | **181** |
| `scripts/eval-lint.sh` | PASS | **PASS** (should_trigger=6, should_not_trigger=5, capability=1) |

## What went

- **The `Triggers:` list.** Seven quoted phrases in EN/ES/CA, 400+ chars sitting in context on every turn the skill is installed whether or not it fires. The model matches by meaning; the list was pure per-turn cost. The description now leads with what it does, keeps the modality list that actually discriminates, and closes with **two** real boundaries — `NOT still-image generation/editing (that is replicate-images); NOT code-rendered React compositing (that is remotion-video)` — where before there was one.
- **`## The one rule`.** A standalone section restating what the pipeline's own step 1 and the first anti-patterns row already say. The rule is worth keeping, so it moved into the lead paragraph of `Pipeline shape` — the step it governs — with its reasoning (metered re-rolls, mux-time aspect/sample-rate surprises) intact.
- **`## Hand-offs` and `## References` tails.** Trailing indexes whose links all already appear inline at point of use. Verified link survival before deleting: `replicate-images` (img→video input + worked pipeline), `fal`/`replicate` (model selection + cost), `remotion-video` (routing row), `references/models-and-params.md` (×3), `references/ffmpeg-assembly.md` (assembly lead). The single link that lived **only** in the tail — `video-shorts` — was promoted into the "Faceless short" table row, and the ffmpeg reference's caption-burning coverage folded into its inline mention.

## What stayed, and why

- **The anti-patterns table.** Required by the rubric, and a different animal from a rationalization bank: it names concrete failure modes (`-c copy`-concat of mismatched clips, mixing without per-track loudnorm, shipping music without checking rights), not rules already stated above it.
- **All four decision tables** — pipeline shape, TTS tiers, img→video models, music licensing. The flow genuinely branches on each, and the img→video table is the load-bearing "choose by the binding constraint" surface.
- **The bad/good VO sample-rate pair**, which teaches the judgment by demonstration rather than by rule.
- **The end-to-end ordered pipeline** and its `scripts/verify.sh` hook — the actionability anchor.
- **Every dated caveat**: the `eleven_v3` verify-availability-first warning, the 2026-05-07 TTS price cut vs the separate Music cut, the 2026-06 duration/pricing snapshots. Freshness is the heaviest rubric dimension; none of it was touched.

## Invariants checked

- Both `references/` files still linked from the body.
- All `../<sibling>/SKILL.md` links resolve (`video-shorts`, `remotion-video`, `replicate-images`, `fal`, `replicate`).
- Frontmatter parses as YAML; description well under the 1024 hard limit.
- Substance preserved exactly — no recommendation, version, command, or figure changed. `manifest.json` untouched (orchestrator regenerates).

### DIFF
--- a/skills/ai-media/SKILL.md
+++ b/skills/ai-media/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: ai-media
-description: "Use when turning a creative goal into a finished media file by orchestrating multiple generative-media models and gluing the pieces with ffmpeg — narrated explainer, product teaser, faceless short, AI voiceover, image-to-video clip, background score, or muxing/ducking/normalizing/stitching existing assets. Triggers: 'make a narrated video with AI voice and B-roll', 'generate a voiceover and mix it over the footage', 'duck the music under the voiceover', 'stitch the scene clips into one MP4 and normalize loudness', 'turn this still into a 9:16 clip with a music bed', 'genérame una voz en off y un vídeo para este guion con música de fondo', 'posa música de fons i abaixa-la quan parla la veu'. NOT still-image generation/editing (that is replicate-images)."
+description: "Use when a creative goal must become a finished media file: pick and order generative-media models per modality — AI voiceover, image-to-video clips, score — then glue them with ffmpeg (mux, duck, loudnorm, concat). NOT still-image generation/editing (that is `replicate-images`); NOT code-rendered React compositing (that is `remotion-video`)."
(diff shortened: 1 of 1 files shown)


ericrisco:

PR #41, opened 2026-07-25
refactor(aws-essentials): migrate to the Claude-5-era authoring rubric

ericrisco:
Context-engineering pass on `skills/aws-essentials/SKILL.md` per `scripts/skill-migration-brief.md`. No substance changed: every command, version, figure and recommendation is byte-identical to before.

| Metric | Before | After |
|---|---|---|
| `description` chars | 806 | **350** |
| Body bytes | 13184 | **11806** |
| Body lines | 224 | **207** |

## What went

- **The `Triggers:` list** — ten quoted phrases across EN/ES/CA. The description is in context on every turn the skill is installed, invoked or not; the model matches by meaning, so the list was pure per-turn cost. The description now carries what it does, when to reach for it, and the `NOT ... (that is deployment) / NOT ... (that is secure-coding)` boundary — both real siblings.
- **"Operating posture — three rules"** — a rule bank whose three rules were each already stated at the step they govern: Apr-2023 S3 defaults in the S3 section, `Action`/`Resource` scoping in the IAM section and the anti-patterns table, roles-over-keys in the hardening checklist. Its one non-duplicated detail (the long-lived `AKIA…` key in a `.env`) moved next to "default to roles" in the IAM section.
- **"Boundary in one sentence"** — a verbatim restatement of the description's boundary. The `secure-coding` half is now inline in the IAM section, phrased as the distinction a reader needs at that point (cloud IAM vs authorization inside app code).
- **"Cross-links" tail index** — `deployment` and `dynamodb` were already linked at point of use; `postgresdb` moved inline into the RDS section, so no link was lost.

## What stayed, and why

- **The anti-patterns table** (12 rows) — required by the rubric and a different animal from a rationalization bank: concrete failure modes with why-it-bites and a fix, not rules restated.
- **The service decision table** — the flow genuinely branches (S3 / RDS / Fargate vs EC2 / CloudFront / "you may not need AWS").
- **The account zero-day hardening checklist** — a real ordered gate, not prose.
- **Both worked bad/good pairs** (the `"*"`-on-`"*"` policy, the public-read bucket) — they teach judgment by demonstration.
- All CLI blocks, the IAM/trust policy JSON, the Fargate cold-start and utilization figures, the RDS encrypt-at-create constraint, and the OAC-over-OAI guidance — untouched.

## Invariants checked

- `references/iam-least-privilege.md` and `references/rds-cloudfront-recipes.md` are both still linked from the body.
- Every `../<sibling>/SKILL.md` link resolves: `cloudflare`, `vercel`, `deployment`, `secure-coding`, `postgresdb`.
- Frontmatter parses as YAML; `name`/`origin`/`tags`/`recommends` unchanged; `description` well under the 1024 hard limit.
- `manifest.json` deliberately untouched — the orchestrator regenerates it.

### DIFF
--- a/skills/aws-essentials/SKILL.md
+++ b/skills/aws-essentials/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: aws-essentials
-description: "Use when standing up the core AWS surface a small product needs — create a private S3 bucket for uploads, provision an encrypted RDS Postgres, choose ECS Fargate vs EC2 for a container, put CloudFront in front of S3, write or tighten an IAM policy, or harden a fresh account (root MFA, no long-lived keys). Triggers: 'set up an S3 bucket for user uploads', 'this role has AdministratorAccess, scope it down', 'spin up Multi-AZ Postgres on RDS', 'my bucket is public and I don't know why', 'this role can do everything', 'I can't encrypt RDS now', 'Fargate or EC2 for a low-traffic app', 'monta un bucket S3 privado', 'permisos mínims a IAM', 'posa CloudFront davant del bucket'. NOT the CI pipeline that ships your container (that is deployment), NOT app-code access-control review (that is secure-coding)."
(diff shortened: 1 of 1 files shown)


ericrisco:

PR #44, opened 2026-07-25
refactor(analytics): migrate to the Claude-5-era authoring rubric

ericrisco:
Context-engineering pass on `skills/analytics/SKILL.md` per `scripts/skill-migration-brief.md`. No substance changed — no recommendation, version, figure, date, or command was touched.

## Metrics

| | Before | After |
| --- | --- | --- |
| `description` chars | 785 | **344** (target ≤350) |
| SKILL.md bytes | 13408 | **11475** |
| SKILL.md lines | 219 | **197** |
| eval-lint | — | **PASS** (should_trigger=8, should_not_trigger=6, capability=1) |

## What went

- **The `Triggers:` phrase list** — 13 quoted phrases across English, Spanish and Catalan, 441 of the description's 785 chars. The model matches by meaning; a keyword list is per-turn cost for every user who has the skill installed, invoked or not, and buys no discriminative power.
- **`## When to use`** — six bullets restating the description almost verbatim. Cross-layer repetition.
- **`## The one rule that governs everything`** — the rule survives, the pedestal does not. The anti-rename argument now opens Step 1, the step it actually governs. The order-of-work line (`taxonomy → SDK wiring → consent gate → PII scrub → funnel + validation`) stays in the intro because it sequences all five steps, not just one.
- **Three prose sentences after the routing table** — each restated a row of the table immediately above it. The gdpr-privacy boundary in particular was stated three times (description, table, prose) and is stated a fourth time in Step 3, where it is actionable; that instance is the one worth keeping.
- **`## References` tail index** — all three files were already linked inline at point of use with their contents described there.

## What stayed, and why

- **Anti-patterns table** — rubric-required, and it names concrete failure modes (render-fired captures, server/browser `user_id` mismatch, email as GA4 `user_id`) rather than restating rules from above.
- **`## When NOT to use` routing table** — routes to nine siblings, well beyond what a ≤350-char description can carry. This is the skill's scope boundary, not a description echo.
- **GA4 vs PostHog vs both decision table** — the flow genuinely branches here.
- **Bad/good event-naming pairs** — teach judgment by demonstration.
- **`## Verify`** block describing `scripts/verify.sh`; the script itself is untouched.

## Invariants checked

- All three `references/` files still linked from the body: `event-taxonomy.md` (Step 1), `ga4-setup.md` (Steps 2 and 3), `posthog-setup.md` (Step 2). No orphans.
- Every sibling named in the description and routing table verified to exist under `skills/`: `dashboard`, `kpi-framework`, `ab-testing`, `gdpr-privacy`, `reporting`, `clickhouse-analytics`, `duckdb`, `sql`, `observability`, `data-policy`, `forecasting`.
- The four NOT boundaries in the description are exactly the four `route_to` targets the `should_not_trigger` evals assert against, so the description discriminates on the cases that are tested.
- Frontmatter parses as YAML; `name` matches the directory; `origin: risco` intact.
- `manifest.json` deliberately not touched — the orchestrator regenerates it.

Only `skills/analytics/SKILL.md` is in the diff.

### DIFF
--- a/skills/analytics/SKILL.md
+++ b/skills/analytics/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: analytics
-description: "Use when instrumenting product or web analytics — adding GA4 or PostHog to an app, designing an event taxonomy, wiring funnels, or making event collection consent-compliant. Triggers: 'set up GA4', 'install PostHog', 'track signup and checkout', 'design the event taxonomy', 'our purchase events fire twice / double-counting', 'PII leaking into event props', 'Consent Mode v2 broke our Google Ads conversions', 'missing client_id / distinct_id', 'instrumentar analítica de producto', 'añade GA4 / PostHog', 'los eventos se cuentan dos veces', 'consentiment abans de capturar dades'. NOT charting the captured data (that is dashboard), NOT choosing which metrics matter (that is kpi-framework), NOT experiment math (that is ab-testing), NOT the legal cookie text (that is gdpr-privacy)."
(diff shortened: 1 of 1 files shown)


ericrisco:

PR #46, opened 2026-07-25
refactor(angular): migrate to the Claude-5-era authoring rubric

ericrisco:
Context-engineering pass on `skills/angular/SKILL.md` per `scripts/skill-migration-brief.md`. No substance changed — no recommendation, version, command, figure, or API claim was altered.

## Numbers

| | Before | After |
|---|---|---|
| description chars | 734 | **346** |
| body bytes | 14907 | **13805** |
| body lines | 288 | **275** |

## What went

- **The `Triggers:` list in the description** — 12 phrases across English, Spanish and Catalan. It is in context on every turn the skill is installed, invoked or not, and the model matches by meaning anyway. Pure per-turn cost, zero discriminative power.
- **`## When to use`** (8 lines) — a restatement of the description in list form.
- **`## References`** tail index — folded into inline mentions. Both files keep a link, now at their point of use:
  - `references/migration.md` → the brownfield row of "Decide first"
  - `references/signals-rxjs.md` → the RxJS rule in "Data fetching"

  Each inline mention absorbed the contents summary the index used to carry, so nothing about what is in those files was lost.
- **`## When NOT to use`** → compressed to `## Not this skill`, prose instead of 7 bullets. Every sibling route survives, and the bare ids (`vue-nuxt`, `svelte`, `solid-js`, `nodejs`, `testing-web`, `e2e-testing`, `react`) are now real `../<id>/SKILL.md` links. All of them resolve.

## What stayed, and why

- **The anti-patterns table.** The rubric requires one, and it names 9 concrete failure modes (`effect`-to-derive loops, `@for` without `track`, `detectChanges()` as a stale-view band-aid) rather than restating rules from above it.
- **The "Decide first" decision table.** The flow genuinely branches three ways: greenfield, brownfield, "view not updating".
- **Every `Rule: … Why: …` bullet.** These are not a rule bank in an appendix — each one sits next to the step it governs and explains why it is absolute. That is the shape the brief asks rule banks to be refactored *into*.
- **All six Bad → Good pairs.** They teach judgment by demonstration.

The description keeps the full Angular-specific API surface (standalone components, signals, zoneless change detection, `@if`/`@for`/`@defer`, `inject()` DI, `resource()`/`httpResource()`, RxJS interop, NgRx SignalStore, ng CLI) and all three `NOT … (that is <sibling>)` boundaries against `react`, `nextjs`, and `typescript` — the discrimination test is intact.

## Verification

- `bash scripts/eval-lint.sh | grep '^angular'` → PASS
- description 346 chars, parses as single-line quoted YAML
- both `references/` files linked from the body
- every `../<sibling>/SKILL.md` link resolves under `skills/`
- `manifest.json` untouched (orchestrator regenerates); `npm test` / `npm run validate` not run — no `node_modules` in this worktree, per the brief

### DIFF
--- a/skills/angular/SKILL.md
+++ b/skills/angular/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: angular
-description: "Use when building, refactoring, reviewing, or debugging an Angular app (Angular 20/21+: standalone components, signals, zoneless change detection, the @if/@for/@defer built-in control flow, inject() DI, resource()/httpResource(), RxJS interop, NgRx SignalStore, the ng CLI). Triggers: 'why isn't my view updating', migrating NgModules to standalone, replacing *ngFor with @for, decorator inputs vs input() signals, 'monta una app Angular', 'componente standalone con signals', 'com passo a zoneless', going zoneless / removing Zone.js, app.config.ts providers, toSignal/toObservable interop, ng generate/ng update. NOT React (that is react), NOT Next.js (that is nextjs), NOT a plain TypeScript language question (that is typescript)."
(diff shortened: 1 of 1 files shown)


ericrisco:

PR #50, opened 2026-07-25
refactor(brand-identity): migrate to the Claude-5-era authoring rubric

ericrisco:
Context-engineering pass on `skills/brand-identity/SKILL.md` per `scripts/skill-migration-brief.md`. No substance changed: same recommendations, versions, figures, and commands.

| Metric | Before | After |
| --- | --- | --- |
| `description` chars | 795 | **338** |
| Body bytes | 12887 | **10622** |
| Body lines | 161 | **135** |
| Hard directives (MUST/NEVER/ALWAYS) | 5 | **0 shouted; 2 kept as stated absolutes with their reason** |

## Description

Dropped the eight-phrase `Triggers:` list — keyword variants in three languages that the model already matches by meaning, and that are in context on every turn the skill is installed, invoked or not. Dropped the `press-kit` clause too: it did not discriminate, and the route still lives in the body's hand-off table.

Kept the two boundaries that let a reader choose this skill over its nearest siblings — `NOT the applied UI pixels (that is design), NOT the words or tone (that is brand-voice)` — both verified real under `skills/`.

## Removed

- **"When to use / When NOT to use"** — the Use-when bullets restated the description, and its five-row routing table duplicated the trailing "Hand-off" list. Merged both into a single **Boundary + hand-off** table at the end. All five routes survive (`brand-voice`, `design`, `marketing`, `press-kit`, `presentations`), with the `design` hand-off detail folded into its row.
- **"The deliverable contract (four parts)"** — a rule bank restating the four sections that immediately followed it; each MUST-bar and its *Why* was already present in the section it governed, or in the anti-patterns table. Replaced by one line naming the four parts and the "a part missing its bar is incomplete, not lite" standard.
- **The five shouted MUSTs.** Where the model reaches the same conclusion unaided (variation set, clear space, minimum size, 2–3 typefaces), the requirement now simply reads as the instruction.

## Kept — deliberately

- **The anti-patterns table, whole.** Concrete misuse modes ("stretch the logo to fill the space", "this blue is close enough"), not rules restated from above. Only the column header changed, `Rationalization` → `Misuse in the wild`, to shed the borrowed-catalog convention flagged by rubric dimension 7.
- **Two absolutes, each stated next to the step it governs, with the reason it is absolute:**
  - four color channels — HEX-only silently breaks print (no CMYK) and wide-gamut screens (no OKLCH), the two places nobody tests;
  - WCAG 2 AA 4.5:1 / 3:1 — a conformance threshold, not a taste call.
  - The tokens export is stated the same way: ship it even when only a PDF was asked for, because it is the only artifact `design` can read.
- The favicon/format matrix, the worked `design-tokens.json` + CSS mapping, the W3C Design Tokens 2025.10 (2025-10-28) grounding, the WCAG 1.4.3 logo-text caveat, and the **Verify** block for the existing `scripts/verify.sh`.
- Both `references/` files linked inline at point of use — `logo-and-assets.md` and `color-and-tokens.md`.

`manifest.json` untouched; only `skills/brand-identity/SKILL.md` is in the diff.

### DIFF
--- a/skills/brand-identity/SKILL.md
+++ b/skills/brand-identity/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: brand-identity
-description: "Use when defining a project's visual identity foundation from scratch or consolidating ad-hoc brand assets into one system: a logo brief, a color system, a type system, usage rules, and an exported design-tokens.json that downstream skills consume. Use when there is no visual foundation yet, or a rebrand must unify scattered colors/fonts. Triggers: 'build our brand identity', 'create a brand book / brand guidelines', 'logo brief', 'pick our color palette', 'define our type system', 'consolidate our random fonts and hex codes scattered across files into one system', 'necesito una identidad de marca', 'crea la guia de marca amb les normes d'us del logo'. NOT the applied UI/pixels (that is design), NOT the words/tone (that is brand-voice), NOT a journalist asset pack (that is press-kit)."
+description: "Use when a project needs its visual foundation built or consolidated into one system: logo brief, color system in HEX/RGB/CMYK/OKLCH with proven AA contrast, type system, usage rules, and an exported W3C design-tokens.json that later skills consume. NOT the applied UI pixels (that is design), NOT the words or tone (that is brand-voice)."
 tags: [brand, identity, logo, color, typography]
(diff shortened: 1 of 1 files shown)


ericrisco:

PR #55, opened 2026-07-25
refactor(author-skill): teach the Claude-5-era rubric instead of the old one

ericrisco:
Migrates `author-skill` to `scripts/skill-rubric.md` per `scripts/skill-migration-brief.md`.

This is the meta skill, so it *documented* the conventions the migration removes — trigger lists, a 120-line body floor, rationalization tables. Changing that guidance **is** the migration here; the skill's own shape was brought in line at the same time.

| | Before | After |
|---|---|---|
| `description` | 962 chars | **346 chars** |
| `SKILL.md` | 18394 bytes / 226 lines | **16698 bytes / 193 lines** |
| `eval-lint.sh` | — | **PASS** (7 / 5 / 2) |

## Description

Dropped a 15-phrase `Triggers:` bank spanning three languages plus the "Knows the rsc conventions: hybrid layout, tags/recommends, manifest.json, the CLI, the Knowledge map" clause. Kept the `Use when` lead, the verbs that discriminate, and both boundaries — `NOT building a product feature (that is specify)` and `NOT designing an agent loop (that is building-agents)`. Both siblings verified under `skills/`.

## What went

- **"When to use / when NOT to use"** — restated the description. Its four delegations survive as one inline line, so every sibling link (`specify`, `plan`, `building-agents`, `harness`, `init`) is preserved.
- **The L0–L3 accompaniment ladder** — `init` owns the dial; re-teaching it here is how the two drift apart. Now one line: read the profile, work at its level, point at `init`.
- **"See Also"** — a tail index whose every link already appears inline at point of use, including all three `references/` files.
- **The rationalization framing** of the anti-patterns table.

## What changed (the guidance itself)

- Description rule 2 was **"Trigger-rich"** and mandated a `Triggers:` list with a non-English phrasing. It is now **"Discriminative, not exhaustive"**, explains why a keyword bank buys no routing accuracy when the model matches on meaning, and states the discrimination test.
- Rule 4 was `≤ 1024 characters`; now **aim ≤ 350, 1024 is the hard limit**.
- **"Keep the body ~120–400 lines"** was a floor. Now: *the smallest body that still routes correctly; 400 is a ceiling, there is no floor.*
- **"Match the form to the failure"** prescribed a *"Prohibition + rationalization table"* for discipline failures. It now prescribes the rule stated at the step it governs, with its why, plus one anti-patterns row — and says why an appendix bank gets skimmed.
- `references/description-recipe.md`: the four-move shape drops move 3 (the `Triggers:` list), gains a *discrimination, not coverage* test, and its worked after-example is now a ~300-char description with no phrase bank.
- `evals/cases.yaml` + `evals/README.md`: the `must_include` rubrics graded a produced skill against the **old** conventions (a `Triggers` list with a non-English phrasing, a 120–400 line body). Left alone they would have kept teaching what this PR removes. Now they grade against the new bar. **Case counts untouched (7/5/2).**

## What stayed, deliberately

- **The anti-patterns table**, rewritten as concrete failure modes ("a reference nothing links to", "`verify.sh` on a process skill") rather than pre-refuted excuses. The rubric requires one, and these are not rules restated from the flow above.
- The **body-vs-references decision table** (a real branch), the four-row **form-to-failure table**, the anatomy tree, the **8-step workflow**, the rsc voice rules, the **orientation footer verbatim**, the **two-ship-gates block** with its behavior-eval commands, and the 02-DOCS grounding section.

No recommendation, version, command, figure, or claim was changed. `manifest.json` untouched. All three `references/` files are still linked inline from the body — verified, zero orphans.

### DIFF
--- a/skills/author-skill/SKILL.md
+++ b/skills/author-skill/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: author-skill
-description: "Use when authoring a NEW skill or editing an existing one for the rsc catalog — writing or fixing a SKILL.md, crafting a trigger-rich third-person description that fires on the right prompts, deciding what goes in references/ vs the body, writing evals/cases.yaml + evals/README.md, or wiring a skill into the catalog (tags, recommends, manifest). Triggers: 'write a skill', 'author a skill', 'create a new skill', 'escribe una skill', 'haz una skill nueva', 'edit this skill', 'improve this skill', 'my skill never triggers', 'my skill always fires', 'fix the description', 'the frontmatter is invalid', 'write evals for this skill', 'add this skill to the catalog', 'audit this skill against the rubric', 'progressive disclosure', 'SKILL.md too long'. Knows the rsc conventions: hybrid layout, tags/recommends frontmatter, manifest.json, the npx @ericrisco/rsc CLI, the Knowledge map. NOT building runtime features (that is the SDD chain) and NOT general docs."
+description: "Use when authoring a NEW rsc skill or editing an existing one — scoping it to one job, writing the description that decides whether it ever loads, splitting the body into references/, writing its evals, auditing it against the rubric. NOT building a product feature (that is `specify`) and NOT designing an agent loop (that is `building-agents`)."
 tags: [skill, authoring, meta]
 recommends: []
 origin: risco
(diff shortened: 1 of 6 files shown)


ericrisco:

PR #56, opened 2026-07-25
refactor(clickhouse-analytics): migrate to the Claude-5-era authoring rubric

ericrisco:
## Metrics

| | Before | After |
|---|---|---|
| `description` chars | 877 | **350** |
| SKILL.md bytes | 12198 | **11599** |
| SKILL.md lines | 165 | **159** |
| `eval-lint` | PASS | **PASS** (7 should_trigger / 5 should_not_trigger / 1 capability) |

## What went

- **The `Triggers:` phrase list** (8 quoted prompts across EN/ES/CA) — 527 of the 877 description chars. The description is in context on every turn the skill is installed, invoked or not, so keyword lists are pure per-turn cost; the model matches by meaning, and those exact intents already live in `evals/cases.yaml` where they cost nothing at runtime.
- **The five-line pre-DDL fork block**, collapsed to one line. Every signal survives verbatim (laptop/in-process/no-concurrent-writers; CRUD/point-updates/FK/row-locks/RLS/migrations; `clickhouse-server`/replication/100M+ rows/s), and the bold sibling names are now real `../duckdb/SKILL.md` and `../postgresdb/SKILL.md` links.
- **"Each rule, then the one-line reason."** — a signpost restating the structure directly below it.

## What stayed, and why

This skill already met most of the rubric at 165 lines, so the honest diff is small.

- **The anti-patterns table** — required by the rubric, and it names concrete failure modes rather than restating rules.
- **The engine decision table** — the flow genuinely branches on it, and the choice is irreversible without a rebuild.
- **The five schema rules** — each already sits next to the DDL it governs and states *why* inline, not in an appendix; this is the shape the rubric asks for, not a rule bank.
- **Bad/good ingestion and MV pairs** — they teach judgment by demonstration.
- **Query-optimization ladder, Operations, and the `verify.sh` contract** — all actionable and unduplicated.

## Invariants checked

- All three `references/` files (`schema-and-engines.md`, `ingestion-and-mvs.md`, `query-optimization.md`) remain linked inline from the body.
- All five `../<sibling>/SKILL.md` links resolve to real skills: `duckdb`, `postgresdb`, `analytics`, `dashboard`, `kpi-framework`.
- `description` parses as single-line quoted YAML, 350 chars, opens with "Use when", carries an explicit `NOT … (that is <sibling>)` boundary.
- No `manifest.json` edit. No recommendation, version (26.3 LTS / 26.2 / 26.1 defaults), figure, or command changed — this was a context-engineering pass, not a content rewrite.


ericrisco:

PR #58, opened 2026-07-25
refactor(code-review): migrate to the Claude-5-era authoring rubric

ericrisco:
Context-engineering pass on `skills/code-review/SKILL.md` per `scripts/skill-migration-brief.md`. No substance changed — no recommendation, command, threshold or figure was touched.

## Numbers

| | Before | After |
|---|---|---|
| `description` chars | 854 | **349** |
| SKILL.md bytes | 10905 | **9302** |
| SKILL.md lines | 143 | **130** |
| `eval-lint.sh` | — | **PASS** (should_trigger=7, should_not_trigger=6, capability=1) |

## Description

Dropped the ten-phrase trilingual `Triggers:` list and the effort-dial recap. That list is in context on every turn the skill is installed, invoked or not, and the body already instructs semantic matching. What replaces it is the rubric shape — what it judges, when (no SDD chain to key off), and one explicit NOT boundary:

> NOT the SDD gate keyed to 02-DOCS/wiki/sdd/ that also processes incoming review comments (that is `review`).

`review` is the most likely mis-route in the catalog, so the boundary names *both* halves of what it owns and this skill does not: the `02-DOCS/wiki/sdd/` chain and the receiving/incoming-comment loop. From the description alone a reader can tell which one they want: naked diff or third-party PR → here; mid-SDD or answering reviewer comments → `review`.

## What went

- **The "one-line disambiguator" blockquote** — restated the description's NOT boundary. The `../review/SKILL.md` routing pointer survives, folded into the opening paragraph.
- **The trailing "See Also" index** — every link that mattered is now at its point of use: `verify`/`debug`/`analyze` as a single delegation line under the pass table, `secure-coding` at pass 4, `ship` at the verdict, `review` in the intro. The five stack-idiom links went: a generic pointer list, not substance, none in `recommends`.
- **Three anti-pattern rows that restated rules already above them** — "this looks like a bug" (the Verify bad/good pair), "more findings = more thorough" (the north-star paragraph), "no PR description, so I'll guess silently" (now one clause inside the intent paragraph, where it is actually read rather than skimmed in an appendix).

## What stayed, and why

- **The anti-patterns table**, trimmed to four rows that name concrete failure modes rather than repeat rules: green-means-correct, preference-as-finding, skimming a dependency bump, blanket `--fix`.
- **The six-pass order table** — the flow genuinely branches per pass.
- **The false-positive skip-list, the 80% confidence floor, the severity ladder, and the worked where/why/repro/fix example** — these teach judgment by demonstration; they are the skill.
- **`references/pr-workflow.md`** stays linked inline from the comment-emitting section (the only file under `references/`).

## Verify

- `bash scripts/eval-lint.sh` → `code-review PASS`
- All six `../<sibling>/SKILL.md` links resolve (`review`, `secure-coding`, `verify`, `debug`, `analyze`, `ship`).
- `manifest.json` untouched; `npm test` / `npm run validate` left to the orchestrator's global run.

### DIFF
--- a/skills/code-review/SKILL.md
+++ b/skills/code-review/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: code-review
-description: "Use when you have a concrete diff, branch, or GitHub PR to judge on its own merits and there is no rsc-SDD spec/plan/constitution chain to key off — the standalone, spec-less giving pass behind the executable /code-review. Triggers: 'review this PR before I merge', 'revisa este diff', 'revísame el PR', 'què està malament en aquest canvi', 'tear apart the contractor's branch', 'is this PR safe to merge', 'is this dependency bump safe', '/code-review --comment 1432', 'high-signal review of this change'. Scales coverage by effort (low/medium = fewer high-confidence findings; high/max = broader, may surface uncertain ones labelled [question]); read-only by default, posts comments only with --comment and edits only with --fix. NOT the SDD review gate keyed to 02-DOCS/wiki/sdd/ that also processes incoming comments and pushes back (that is review)."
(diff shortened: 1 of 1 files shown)


ericrisco:

PR #60, opened 2026-07-25
refactor(clarify): cut the description 930→343 chars and fold the rule banks into the pass

ericrisco:
Context-engineering pass on `clarify` against `scripts/skill-rubric.md`, following `scripts/skill-migration-brief.md`. Phase semantics and chain position are untouched — this is a format pass, not a content rewrite.

| Metric | Before | After |
| --- | --- | --- |
| `description` | 930 chars | **343 chars** (-63%) |
| Body | 15352 B / 163 lines | **12280 B / 122 lines** (-20%) |

## What went

- **The `Triggers: '…'` list** (10 phrases, EN+ES). The model matches by meaning; on a `core`-profile skill that list is paid for on every turn it is installed. The description was rebuilt as what/when plus a discriminative `NOT` boundary against the real neighbours: `specify` (writes the spec clarify interrogates), `plan` (designs the build from it), `analyze` (the cross-artifact check people confuse with this single-spec one).
- **"When to use / when NOT to use"** (17 lines) — the description carries that boundary now. Its one non-restating line ("specify creates, clarify de-risks, plan designs") moved into the opening chain-position paragraph; the `debug` off-ramp moved to *Next in the chain*, where routing already lives.
- **The model-routing mechanics paragraph** — `sdd` owns that machinery. Now one line naming this phase's tier (`balanced`) plus a pointer to `../sdd/references/model-routing.md`.
- **The standalone "Question-shape rule" section** — its four bullets govern step 4 of the pass, so they now sit inside step 4. A rule in context is followed; a rule in its own section gets skimmed and then re-stated by the anti-patterns table below it.
- **Two of eight anti-pattern rows** — "the user is busy, I'll guess" duplicated the silent-default row, "one question at a time" duplicated the ask-everything row. Folded into the rows they duplicated (8 → 6).
- **The 8-item "Done check"** — it restated the five steps line for line. A gate phase still needs a stated pass condition, so it became a three-line **Exit gate** paragraph with the same conditions.
- **The "See Also" tail index** — every sibling it named is already named inline at the point it matters, and the `../harness/SKILL.md` link survives inline next to the dial it explains.

## What stayed, deliberately

- **The ten-row ambiguity taxonomy** and **the five-step pass** — the substance of the phase; the capability eval asserts both.
- **The worked profile-photo micro-example** — teaches the inventory → resolve → rank judgment by demonstration, which no rule bank does.
- **An anti-patterns table** (the rubric requires one), reframed from rationalization→reality to anti-pattern→why it breaks the gate.
- **The accompaniment dial**, compressed to a pointer + an L0..L3 table: the L-level mapping here is *this phase's* question-count contract (asserted by the capability eval), not a restatement of what `init` owns.
- **The Orientación block** and the "no new artifact — edit the spec in place" contract, unchanged.
- **No result-envelope block was added.** `clarify` never had one; adding it would be a content change.

## Verification

- `bash scripts/eval-lint.sh | grep '^clarify'` → `PASS (should_trigger=7, should_not_trigger=5, capability=1)`.
- `description` is 343 chars and parses as single-line quoted YAML.
- No `references/` dir under `skills/clarify/`, so nothing is left unlinked. All `../<sibling>/` links (`sdd`, `sdd/references/model-routing.md`, `harness`) resolve.

### DIFF
--- a/skills/clarify/SKILL.md
+++ b/skills/clarify/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: clarify
-description: "Use when a spec exists and you need to de-risk it BEFORE planning — hunt the ambiguities, unstated assumptions, edge cases, and underspecified areas, ask the user the high-leverage questions, then bake the answers back into the spec. The fourth phase of the rsc SDD chain (constitution → specify → clarify → plan → tasks → analyze → implement → verify → review → ship). Triggers: 'clarify the spec', 'what's ambiguous here', 'poke holes in this spec', 'is this spec ready to plan', 'aclara el spec', 'qué falta en este spec', 'review the requirements before we build', 'de-risk before planning', 'what questions should I answer first', 'this spec feels vague', a spec under 02-DOCS/wiki/sdd/specs/ heading into planning, or a plan that keeps stalling on unknowns. Reads the spec + constitution + harness profile; writes resolved answers back into the same spec file. NOT spec authoring (specify) and NOT the technical plan (plan)."
+description: "Use when a spec exists and must be de-risked before planning — hunt its ambiguities, unstated assumptions and edge cases, ask the few build-changing questions, bake the answers back into the spec. The rsc SDD gate between `specify` (writes the spec) and `plan` (designs the build). NOT the cross-artifact consistency check (that is `analyze`)."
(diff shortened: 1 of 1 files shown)


ericrisco:

PR #63, opened 2026-07-25
refactor(chatbot): migrate to the Claude-5-era authoring rubric

ericrisco:
Context-engineering pass on `skills/chatbot/SKILL.md` per `scripts/skill-migration-brief.md`. No content rewrite: no figure, source, date, command or recommendation changed.

| Metric | Before | After |
| --- | --- | --- |
| description | 825 chars | **350 chars** |
| body | 14810 bytes / 168 lines | **13364 bytes / 151 lines** |

## Description

Dropped the seven-phrase `Triggers:` list (EN + ES/CA). It is in context on every turn the skill is installed, invoked or not, and the model matches by meaning — the body already said so. What is left is the rubric shape: what it does, when, and an explicit NOT boundary naming two real siblings, `building-agents` (the loop/index under the bot) and `customer-support` (a human on one ticket). A reader can pick between the three from the description alone.

## What went

- **`## When to use / When NOT to use`** — the "Use when" paragraph restated the description clause for clause, and the NOT bullets restated the intro paragraph's building-agents / customer-support delegation a third time. Compressed into one routing line; every sibling target survived (`rag`, `prompt-engineering`, `agent-safety`, `agent-eval`, `retention`, `client-onboarding`, `automation-flows`, `whatsapp-telegram`).
- **`## References` tail index** — both reference files were already linked inline where they are needed. The index's only unique content was its per-file blurb, which moved into those inline mentions; Layer 3's dangling "full checklist in the reference" became the actual link. Invariant holds: both `references/` files are still linked from the body.

## What stayed, on purpose

- The **anti-patterns table** (9 rows) — concrete failure modes with sources, a different animal from a rule bank.
- **"The one rule"** and the Air Canada liability case — the skill's thesis, and it states *why* it is absolute rather than shouting.
- The **four-layer spine**, the **forbidden-topic bucket table**, the **handoff trigger table**, the **launch-metric benchmarks + rollout ladder + kill switch**, and the **bad/good system-prompt pair**: real branch points and judgment taught by demonstration.

`evals/`, `scripts/verify.sh` and `references/` untouched. `manifest.json` not touched (orchestrator regenerates).

### DIFF
--- a/skills/chatbot/SKILL.md
+++ b/skills/chatbot/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: chatbot
-description: "Use when putting a support or sales bot on a live website and it must behave — writing or hardening its persona/system prompt, designing the bot→human handoff, stopping it inventing prices/policies/refunds, defending against jailbreaks and prompt injection, qualifying leads, or setting launch metrics and a kill switch. Triggers: 'put a chatbot on our site that answers from our docs', 'add a chat widget', 'our bot promised a refund we don't offer — lock it down', 'stop people jailbreaking our website assistant', 'design the handoff to a human agent', 'qualify leads and book demos', 'necesito un chatbot de ventas para la web que califique leads i reservi demos'. NOT building the agent loop/RAG index/eval harness under it (that is building-agents), and NOT a human answering one live ticket (that is customer-support)."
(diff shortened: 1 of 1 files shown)


ericrisco:

PR #66, opened 2026-07-25
refactor(cloudflare): migrate to the Claude-5-era authoring rubric

ericrisco:
Context-engineering pass on `skills/cloudflare/SKILL.md` per `scripts/skill-migration-brief.md`. No substance changed: every command, figure, version and pricing claim is byte-identical.

## Numbers

| | Before | After |
|---|---|---|
| `description` | 741 chars | **350 chars** |
| body | 11021 bytes / 191 lines | **10336 bytes / 186 lines** |
| eval-lint | PASS | **PASS** (should_trigger=7, should_not_trigger=5, capability=1) |

Diff is 3 insertions / 8 deletions across one file.

## What went

- **The `Triggers:` phrase list** (eight phrases across EN/ES/CA). The description is in context on every turn the skill is installed, invoked or not; a keyword list is pure per-turn cost when the model already matches by meaning. This alone is the bulk of the 391-char saving.
- **The `## References` tail index.** Both reference files were already linked at point of use — `storage-primitives.md` from the storage decision table, `wrangler-config.md` from the config-anatomy and Static Assets sections — so the index was a second copy of links the reader had already passed. The three details that existed *only* in the index were folded into those inline pointers rather than dropped: consistency semantics, the full limits/pricing tables and Hyperdrive for external Postgres onto the storage pointer; asset-routing rules onto the Pages→Workers pointer.
- Two description redundancies: `wrangler.jsonc/wrangler.toml` → `wrangler.jsonc` (the body states all three config filenames are accepted) and `a Vite/React SPA` → `SPA`.

## What stayed, deliberately

- **The anti-patterns table** — 8 concrete failure modes (stale KV reads, D1 past 10 GB, missing `compatibility_date`, avoiding R2 over non-existent egress fees, Workers Sites, secrets in `vars`, D1-as-pooled-connection, subrequest fan-out). Required by the rubric and not a restatement of anything above it.
- **The storage-primitive decision table** — this is the branch that shapes the whole architecture; it is exactly the case the rubric keeps decision tables for.
- **"Limits that reshape your design"** and the Workers Paid plan note — architecture inputs, and the half of the description that routes "I hit a runtime limit" prompts.
- The one-paragraph binding mental model, the `wrangler.jsonc` anatomy, the typed `Env` example, the secrets/`.dev.vars` rules and the Queues producer/consumer wiring. All load-bearing and all already at point of use.

## Gates

- `name` matches the directory id; `origin: risco` present; frontmatter parses as YAML with the description on a single quoted line.
- All three siblings named in the `NOT` boundary exist (`skills/deployment`, `skills/nextjs`, `skills/domains-dns`), as do all six `recommends` ids.
- Both files under `skills/cloudflare/references/` are still linked from the body.
- No `../<sibling>/SKILL.md` links in the body to resolve.
- `manifest.json` deliberately untouched — the orchestrator regenerates it.

### DIFF
--- a/skills/cloudflare/SKILL.md
+++ b/skills/cloudflare/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: cloudflare
-description: "Use when working on Cloudflare's edge platform — writing or editing wrangler.jsonc/wrangler.toml bindings, choosing between D1/KV/R2/Durable Objects/Queues, deploying a Worker or a Vite/React SPA via Static Assets, wiring Queue producers/consumers, or hitting a Cloudflare runtime limit and needing the architecture fix. Triggers: 'deploy to Cloudflare Workers', 'add an R2 bucket binding', 'D1 vs KV for sessions', 'why is my KV read stale after writing', 'too many subrequests in my Worker', 'wrangler deploy', 'desplegar a Cloudflare Workers', 'quin binding faig servir per R2'. NOT generic CI/release deploy (that is deployment), NOT the Next.js framework wiring itself (that is nextjs), NOT Cloudflare DNS records (that is domains-dns)."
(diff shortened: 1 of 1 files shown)


ericrisco:

PR #67, opened 2026-07-25
refactor(competitor-watch): migrate to the Claude-5-era authoring rubric

ericrisco:
Context-engineering pass on `skills/competitor-watch/` per `scripts/skill-migration-brief.md`. No substance changed — no recommendation, figure, version, command, path or source date differs.

| Metric | Before | After |
|---|---|---|
| `description` | 943 chars | **323 chars** |
| Body | 14007 bytes / 199 lines | **12587 bytes / 184 lines** |
| `scripts/eval-lint.sh` | — | **PASS** (should_trigger=5, should_not_trigger=5, capability=1) |

## What went

- **The `Triggers:` list.** Eight quoted phrases across English, Catalan and Spanish. The description is in context on every turn the skill is installed, invoked or not, and the model matches by meaning — the keyword list was pure per-turn cost.
- **Two of the four `NOT` boundaries** (`pricing`, `sales-pipeline`). Both are already rows in the body's Handoffs table. The description keeps the discriminating pair a reader actually needs to choose between siblings: `market-research` (who the rivals are) and `data-scraper` (a one-off extraction with no cadence).
- **The "Three hard stops up front" bullet list.** All three routes appear verbatim in the Handoffs table — the same rule paid for twice, in two places that can drift apart.
- **The three bullets under "The loop is the product".** "Not a one-shot study" and "watching 30 competitors" are already anti-pattern rows; "append, never overwrite" is already step 5 of the change loop. The one claim only those bullets carried — the vital-few **3–7** — moved next to the step it governs (Ground & scope, step 2), and the sourced **Orient → Gather → Analyze → Report → Act** cycle merged into the opening paragraph with its citation intact.

## What stayed, on purpose

- **The ethics gate**, in full. It reads like a rule bank but it is not one: it runs before any capture, each rule states *why* it is absolute (front-page test, Code-of-Ethics violation, the access-control bright line), and it ends in a concrete refuse-and-reframe. Moving it to an appendix would mean it gets skimmed.
- **The surface → cadence → selector table** — the flow genuinely branches per axis, and the 5–15 min / hourly / daily tiering is the actionable core of the skill.
- **The anti-patterns table** (rubric-required), **the Handoffs table**, the change loop, both worked bad/good pairs, the tracker artifact contract, and the sourced tooling numbers (changedetection.io, Crayon/Klue/Kompyte/Visualping/ChangeTower, Wayback's archive-not-monitor caveat).

## Invariants checked

- Both `references/` files still linked inline from the body: `tracker-schema.md`, `monitoring-config.md`.
- All seven `../<sibling>/SKILL.md` paths resolve to real skills.
- Frontmatter parses as YAML; `name` matches the directory id; `origin: risco` intact; description well under the 1024 hard limit.

### DIFF
--- a/skills/competitor-watch/SKILL.md
+++ b/skills/competitor-watch/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: competitor-watch
-description: "Use when you keep a named set of rivals under continuous, dated observation — tracking their pricing, features, positioning, and changelog over time, turning page diffs into a maintained tracker plus a classified change log on a cadence. Use when asked what moved at a competitor this month, whether a rival changed price or shipped something, or for a feature matrix kept current rather than written once. Triggers: 'set up competitor monitoring', 'watch their pricing page', 'did [rival] reposition or launch anything', 'what moved at [competitor] this month', 'keep this comparison matrix updated', 'track how their positioning shifted over 6 months', 'monitoritza els competidors i avisa-m quan canviïn de preu', 'vigila los precios de la competencia'. NOT sizing the market or producing the competitor list (that is market-research), NOT setting our own price (that is pricing), NOT building the sales battlecard (that is sales-pipeline)."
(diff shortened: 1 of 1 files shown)


ericrisco:

PR #69, opened 2026-07-25
refactor(building-agents): migrate to the Claude-5-era authoring rubric

ericrisco:
Context-engineering pass on `skills/building-agents/SKILL.md` per `scripts/skill-migration-brief.md`. No substance changed — no recommendation, version, figure, command, or claim differs.

## Numbers

| | Before | After |
|---|---|---|
| `description` chars | 713 | **346** (target ≤350) |
| body bytes | 25,032 | **19,908** (−20%) |
| body lines | 469 | **386** (ceiling 400) |
| eval-lint | PASS | **PASS** (7 / 7 / 2) |

## Description

Dropped the ten-item `Triggers: '…'` keyword list and the trailing `FastAPI/Python, Next.js, Go, Flutter, Postgres stacks.` enumeration — in context on every turn the skill is installed, and the model matches by meaning anyway. Replaced with what it does · when · an explicit boundary against two verified siblings:

> Use when building or restructuring an LLM agent — provider adapter, tool calling, structured output, RAG, agent loop, eval gate, cost routing, tracing, MCP server — model-agnostic across OpenAI/Anthropic/Gemini/OSS so a model swap is a config change. NOT vector-store SQL alone (that is `postgresdb`) or service deployment (that is `deployment`).

## What went

- **ASCII architecture diagram** (19 lines) — restated the intro sentence and the section headings in box-drawing characters.
- **"Quick reference" tail index** (14 lines) — all five `references/` files are already linked inline at point of use. Its two unique facts (Gemini `response_json_schema`; orchestrator-worker fan-out with a semaphore) were folded into the adapter preamble and decision rule 2 rather than dropped.
- **The "Use when:" paragraph** — a verbatim restatement of the description.
- **Good/Bad pair 1** (vendor SDK hardwired vs adapter) — the adapter code block sits directly above it and the anti-patterns table already names the failure. The JSON-validation and unbounded-loop pairs stay.
- **`harness` re-teaching in "Project grounding"** — the OKF v0.1 frontmatter/wikilink rules now point at [`wiki-article-template.md`](https://github.com/ericrisco/rsc-harness/blob/main/skills/harness/references/wiki-article-template.md) instead of restating them, so the two cannot drift.
- **The "ECC analogues" line** — five external names, no links, no action.

## What stayed, deliberately

- **Every code block**: provider adapter (OpenAI + Anthropic + `get_provider`), typed tool args and validating dispatch, pgvector RAG with cite-or-refuse, the eval runner with the CI exit code, the OTel `gen_ai.*` span, the FastMCP server. This is the skill; the brief says do not cut technical content to hit a number.
- **The full 12-row anti-patterns table**, retitled from "Rationalization / Reality" to "Anti-pattern / Reality" — the rows name concrete failure modes, which is the table the rubric requires, not the borrowed-urgency kind it warns about.
- **The six decision rules**, the **SDD hand-off gate**, and the **`verify.sh` contract**.
- **Every routing target the evals depend on** — `claude-api`, `harness`, `postgresdb`, `deployment`, prompt-wording, throwaway-prompt — compressed into one hand-off paragraph instead of a bulleted "Do NOT use when" block.

## Checks

- `bash scripts/eval-lint.sh` → `building-agents PASS (should_trigger=7, should_not_trigger=7, capability=2)`.
- All five `references/*.md` are linked from the body, now as markdown links in prose rather than bare paths inside code comments (a path in a code fence is not a link).
- All 12 `../<sibling>/` links resolve to skills that exist.
- `description` is 346 chars and parses as single-line quoted YAML.
- `manifest.json` untouched; only `skills/building-agents/SKILL.md` is in the diff.

### DIFF
--- a/skills/building-agents/SKILL.md
+++ b/skills/building-agents/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: building-agents
-description: "Use when designing or building an LLM agent, tool-using system, RAG pipeline, eval harness, or MCP server in this repo — across any provider (OpenAI, Anthropic, Google Gemini, or OSS via OpenAI-compatible endpoints / litellm). Triggers: 'build an agent', 'add tool calling / function calling', 'structured JSON output', 'RAG / retrieval / embeddings / rerank', 'agent loop / ReAct / orchestrator-worker / multi-agent', 'LLM eval / golden set / LLM-as-judge / regression gate', 'prompt caching / model routing / token budget / cost control', 'trace / observability for LLM calls', 'build an MCP server', or 'make our LLM code provider-agnostic / swap models'. FastAPI/Python, Next.js, Go, Flutter, Postgres stacks."
+description: "Use when building or restructuring an LLM agent — provider adapter, tool calling, structured output, RAG, agent loop, eval gate, cost routing, tracing, MCP server — model-agnostic across OpenAI/Anthropic/Gemini/OSS so a model swap is a config change. NOT vector-store SQL alone (that is `postgresdb`) or service deployment (that is `deployment`)."
 tags: [agents, llm, mcp, rag, evals, ai]
 recommends: [secure-coding, deployment]
 origin: risco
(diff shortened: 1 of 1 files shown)


ericrisco:

PR #74, opened 2026-07-25
refactor(contracts): migrate to Claude-5-era authoring rubric

ericrisco:
Context-engineering pass on `skills/contracts/SKILL.md` against `scripts/skill-rubric.md`. No other file touched; `manifest.json` deliberately left for the orchestrator to regenerate.

| Metric | Before | After |
|---|---|---|
| `description` chars | 750 | **347** (target ≤350, hard limit 1024) |
| Body bytes | 13000 | **11773** (−9.4%) |
| Body lines | 124 | **113** (ceiling 400) |
| `references/` files linked inline | 2/2 | **2/2** |

## What went

- **The `Triggers:` phrase list** (nine quoted phrases across English, Spanish and Catalan) and the "symptoms include …" restatement. That was ~400 chars of the description, in context on **every turn the skill is installed**, invoked or not — and the model matches by meaning, so the list bought nothing the semantic content did not already buy.
- **The "Three rules sit above everything below" bank** at the top of the body. Rule 1 (plain English) was restated by the `Plain-language drafting rules` section and a third time by the anti-patterns table. Rules 2 and 3 carried real substance, so they were **moved next to the step they govern**, not deleted:
  - "state who pays, in one line, beside the clause" now opens `The clauses that allocate risk`, where it is actually applied;
  - the attorney-review rule now lives in `The legal boundary`, where the **ABA Formal Opinion 512 (2024-07-29)** citation and the UPL claim were merged into one bullet instead of appearing twice in two different wordings.
- **The trailing `## References` index.** Both files were already linked inline at point of use, so the index was a second copy of the same two pointers. The `review-playbook.md` inline mention absorbed the MSA↔SOW pointer the index carried, so nothing the index said was lost — and the rubric's invariant (every `references/` file linked from the body) still holds.
- **`(non-negotiable)` from the legal-boundary heading.** The bullet now says *why* the attorney-review line is absolute rather than shouting the label at the reader.

## What deliberately stayed

This is a legal-domain skill, so the bar for cutting was "is this a duplicate?", never "is this long?".

- **Every legal claim, verbatim in substance**: the 12-month-fees liability cap default and the confidentiality/indemnity carve-outs, the fraud / willful-misconduct / bodily-harm exclusions courts strike, the 3–5 year indemnity survival, the **ICC model force-majeure clause (last updated March 2020)**, ESIGN (2000) / UETA (49 states + DC + territories) / eIDAS and the EU's three signature tiers, UPL statutes, and ABA Op. 512 on informed consent before pasting confidential text into public tools.
- **The instrument/side decision table** — the flow genuinely branches on buyer/seller and discloser/recipient, and a wrong guess inverts every default.
- **The seven-row anti-patterns table** — required by the rubric, and a different animal from the rule bank that was removed: it names concrete failure modes, not rules already stated above it.
- **All three bad/good drafting pairs** — they teach judgment by demonstration, including the "any and all claims" pair whose point is a *risk* decision, not a style one.
- The five-step review order, the MSA/SOW reconciliation section, and the six named sibling hand-offs.

No recommendation, figure, date, citation, or version changed. Not a content rewrite.

## Verification

- `bash scripts/eval-lint.sh` → `contracts  PASS (should_trigger=6, should_not_trigger=5, capability=1)`
- `description` 347 chars, parses as single-line quoted YAML; `name` matches the directory id; `origin: risco` intact.
- All 6 `../<sibling>/SKILL.md` links resolve (`terms-conditions`, `e-signature`, `gdpr-privacy`, `proposals`, `ip-trademark`, `invoicing`), as do both `NOT` boundary siblings.
- Both `references/` files still linked from the body.
- `npm test` / `npm run validate` not run — no `node_modules` in this worktree, per the migration brief; the orchestrator runs them globally after merge.

### DIFF
--- a/skills/contracts/SKILL.md
+++ b/skills/contracts/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: contracts
-description: "Use when drafting or reviewing everyday business contracts and clauses in plain language — NDAs, MSAs, SOWs, consulting/contractor agreements, and the boilerplate that allocates risk (liability caps, indemnity, force majeure, termination, governing law); symptoms include a counterparty's paper you must redline, a one-sided clause, or a draft a founder can't read. Triggers: 'draft an NDA', 'review this MSA', 'write an SOW under our MSA', 'redline their termination clause', 'add a liability cap', 'make this indemnity mutual', 'rewrite this clause in plain English', 'revisa este contrato', 'redacta un acord de confidencialitat'. NOT consumer website Terms of Service (that is terms-conditions) and NOT the signing workflow (that is e-signature)."
+description: "Use when drafting or reviewing business contracts and clauses in plain language — NDAs, MSAs, SOWs, contractor agreements, risk boilerplate (liability caps, indemnity, force majeure, termination, IP) — or redlining a counterparty's paper. NOT consumer Terms of Service (that is terms-conditions) and NOT the signing workflow (that is e-signature)."
 tags: [contracts, legal, ndas, msa, sow, clauses, redlining, risk-allocation]
 recommends: [terms-conditions, e-signature, gdpr-privacy, proposals, ip-trademark, invoicing]
 origin: risco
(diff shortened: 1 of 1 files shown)


ericrisco:

PR #75, opened 2026-07-25
refactor(context-budget): migrate to the Claude-5-era authoring rubric

ericrisco:
Context-engineering pass on `skills/context-budget/SKILL.md` per `scripts/skill-migration-brief.md`. No substance changed.

## Numbers

| | Before | After |
|---|---|---|
| `description` chars | 837 | **347** |
| SKILL.md bytes | 11,140 | **9,605** |
| SKILL.md lines | 132 | **114** |
| `eval-lint.sh` | — | **PASS** (should_trigger=5, should_not_trigger=4, capability=1) |

## What went, and why

- **The `Triggers:` list (8 phrases, EN/ES/CA).** The description is resident on every turn the skill is installed, invoked or not, and the model matches by meaning — a keyword list in three languages is pure per-turn cost. Description now carries the rubric shape only: what it does, when, and `NOT dollar spend or caps (that is cost-tracking), NOT finding context via embeddings (that is rag)`.
- **`## When to use this`** — six bullets that restated the description.
- **`## When NOT to use (and where to go)`** — collapsed from eight prose bullets to one paragraph. All six sibling routes survive by name (`cost-tracking`, `rag`, `prompt-engineering`, `building-agents`, `parallel`, `harness`), as do the two distinctions a reader can't derive: cost-tracking owns dollars where this owns attention, and RAG is *how you find* context where this is *how much* you let live.
- **The trailing `## Reference` index.** `references/handoff-and-compaction.md` is already linked inline twice at its two points of use, so the tail was a duplicate pointer. The one detail it named uniquely (the good-vs-bad summary checklist) was folded into the inline mention, so the reference stays fully signposted.

## What stayed, and why

- **The anti-patterns table** — required by the rubric, and these are concrete failure modes ("paste a subagent's full transcript back into the parent"), not a rationalization bank restating rules already given.
- **The "window is filling — what do I do?" decision table** — the flow genuinely branches four ways (Offload / Reduce / Retrieve / Isolate / Hand off).
- **The bad/good worked pairs** under each of the four moves — they teach the judgment call by demonstration, not by adjective.
- **The budget-allocation percentages**, the ~60% compaction threshold, the server-side compaction figures (`compact-2026-01-12`, `compact_20260112`, 150,000 default trigger), and every dated source citation — unchanged.

## Verification

- `bash scripts/eval-lint.sh | grep '^context-budget'` → PASS
- Description 347 chars, single-line quoted YAML, parses clean
- All six `../<sibling>/SKILL.md` links resolve to real directories under `skills/`
- The only file under `references/` is linked from the body (twice, inline)
- `manifest.json` untouched; `npm test` / `npm run validate` left to the orchestrator (no `node_modules` in this worktree)

Only `skills/context-budget/SKILL.md` is touched.

### DIFF
--- a/skills/context-budget/SKILL.md
+++ b/skills/context-budget/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: context-budget
-description: "Use when a long-horizon task is filling the context window and you must decide what to keep, offload, drop, or hand off — large refactors and migrations that won't fit one window, multi-hour research/agent runs, planning a harness that survives across fresh sessions, or deciding when to compact and whether to spawn a subagent. Triggers: 'we're 80% through context, compact before it breaks', 'context is 80% full', 'make this survive a fresh window', 'this conversation is getting huge', 'should I spawn a subagent for this huge log read or just inline it', 'the agent keeps re-reading the same files and forgetting decisions it made earlier', 'se está llenando el contexto, ¿qué guardo y qué tiro?', 'compacta abans que peti'. NOT dollar spend or budget caps (that is cost-tracking), NOT fetching context via embeddings (that is rag)."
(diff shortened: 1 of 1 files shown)


ericrisco:

PR #79, opened 2026-07-25
refactor(data-scraper): trim description to the rubric, drop the tail reference index

ericrisco:
Context-engineering pass on `skills/data-scraper/` per `scripts/skill-migration-brief.md`. No content rewrite: no recommendation, version, figure or command changed.

| Metric | Before | After |
|---|---|---|
| description chars | 679 | 349 |
| body bytes | 13610 | 12524 |
| body lines | 134 | 128 |
| eval-lint | — | PASS (should_trigger=7, should_not_trigger=5, capability=1) |

**Removed**

- The nine-phrase `Triggers: '…'` list (EN/CA/ES) from the description. It sits in context on every turn the skill is installed, invoked or not, and the model matches by meaning anyway — cost with no routing gain.
- The trailing `## References` index. All three reference files are already linked inline at the step that needs them (`legal-compliance.md` in the legal gate, `anti-bot.md` in politeness, `frameworks.md` in resilience), so the index restated links the reader had already passed.

**Kept**

- The legal gate checklist with its case law (Meta v. Bright Data, hiQ v. LinkedIn, Reddit v. Perplexity) — the skill's whole reason to run before the first request.
- The extraction-path and tool-picker decision tables — the flow genuinely branches there.
- The bad/good selector pair, which teaches judgment by demonstration, and the concrete pacing numbers (1 req/1-3s, 2-5 concurrent, backoff with jitter).
- The anti-patterns table, required by the rubric and distinct from a rationalization bank.

**Invariant check**: every file under `references/` is still linked from the body exactly once, inline at point of use. `manifest.json` untouched.


ericrisco:

PR #81, opened 2026-07-25
refactor(dashboard): migrate to the Claude-5-era authoring rubric

ericrisco:
Context-engineering pass on `skills/dashboard/SKILL.md` per `scripts/skill-migration-brief.md`. No substance changed — no recommendation, figure, command, or claim was altered.

## Numbers

| | Before | After |
| --- | --- | --- |
| `description` chars | 761 | **344** |
| SKILL.md bytes | 11,426 | **9,903** |
| SKILL.md lines | 206 | **185** |

`bash scripts/eval-lint.sh` → `dashboard  PASS (should_trigger=7, should_not_trigger=6, capability=1)`

## What went

- **The `Triggers:` list** (7 phrases, EN/ES/CA). It sat in context on every turn the skill is installed, invoked or not, while the model matches by meaning regardless. Replaced with the rubric shape: what it does · when · an explicit `NOT` boundary naming three real siblings.
- **"The one rule: the 5-second read"** — a rule bank that restated the three sections immediately below it. Each claim survives at its point of use (north-star → Tile budget + Layout; comparison → Framing every tile; all three → anti-patterns table). Its one unique line moved into the Decision Test gate, the step it governs.
- **The tail Handoff index** — 5 of its 8 routes already appear inline where the reader needs them. Kept only the 3 that appear nowhere else (ab-testing, design, spreadsheet-ops).
- **The duplicated verify.sh contract**, stated twice. Now once each, split by where the reader needs it: the checks at the run instruction, the required tile keys next to the manifest.
- **The chart kill-list**, compressed from a bulleted block to one line — all four bans and their replacements intact.

## What stayed, deliberately

- The **anti-patterns table** — the rubric requires one, and it names concrete failure modes rather than restating rules already given above it.
- The Decision Test vanity→actionable conversion table and the tiering checklist: the flow genuinely branches at both.
- The chart-shape matrix, and the bad/good pairs for tile framing and layout that teach judgment by demonstration.
- The **full five-tile `dashboard.yaml`** — trimming it would have dropped the example below the 5-9 tile budget the skill itself prescribes.

## Invariants checked

- Both `references/` files (`chart-selection.md`, `tile-schema.md`) still linked inline from the body.
- All 8 `../<sibling>/SKILL.md` links resolve to real skills.
- `description` parses as single-line quoted YAML, 344 chars (≤350 aim, ≤1024 hard limit).
- `manifest.json` untouched; only `skills/dashboard/` in the diff.

### DIFF
--- a/skills/dashboard/SKILL.md
+++ b/skills/dashboard/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: dashboard
-description: "Use when building or fixing a KPI dashboard that decision-makers will actually open and act on within one screen — arranging which metrics earn a tile, framing each tile so it answers a decision at a glance, and picking the right chart per metric. Triggers: 'build a KPI dashboard for leadership', 'exec dashboard on one screen', 'nobody opens the dashboard we built', '20 tiles and I still can't tell if the business is healthy', 'which chart for each of these metrics', 'munta un dashboard de KPIs per a direcció amb el north-star a dalt', 'el dashboard tiene demasiados gráficos y nadie lo mira'. NOT defining the metrics or their targets (that is kpi-framework), wiring the data (that is analytics), or writing a narrative status report (that is reporting)."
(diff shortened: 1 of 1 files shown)


ericrisco:

PR #85, opened 2026-07-25
refactor(course-storytelling): migrate to the Claude-5-era authoring rubric

ericrisco:
Context-engineering pass on `skills/course-storytelling/` per `scripts/skill-migration-brief.md`. No substance changed: same frameworks, same flags, same paths, same numbers.

## Metrics

| | Before | After |
| --- | --- | --- |
| `description` | 1023 chars | **344 chars** (-66%) |
| body | 20607 bytes / 205 lines | **14506 bytes / 137 lines** (-30% / -33%) |
| eval-lint | — | **PASS** (should_trigger=6, should_not_trigger=5, capability=2) |

## Description

Before: a nine-phrase `Triggers:` list plus a five-clause summary of the method, ending on `NOT slide visuals … NOT a content-audit pass` — neither of which is the sibling a reader is actually likely to confuse this with.

After (344 chars):

> Use when lesson or course content is correct but forgettable and a concept has to LAND: profiles the learner, breaks the blocking false belief, then rebuilds it as epiphany story → named model → grounded analogy → proof → so-what. NOT outcomes, assessment or module order (that is `course-builder`), NOT slide visuals (that is `presentations`).

The trigger list went because the model matches by meaning and the body already says so. The boundary now names `course-builder` — the real nearest sibling, and one the body did not mention anywhere: **course-builder decides what the course must prove (outcomes, assessment, module order); course-storytelling makes a concept inside that skeleton land.**

## What went

- **"When to use / When NOT to use"** — restated the description. Its four delegations survive as one dense boundary paragraph in the intro.
- **"The non-negotiables"** (10 numbered rules) — every rule was already stated in the workflow, the recipe, the anti-patterns table or the QA gate. The three carrying reasoning not found elsewhere (break the belief before teaching, never invent proof → `[[NEEDS PROOF]]`, arrived-at not stated) moved next to the step they govern, phrased as *why* rather than NON-NEGOTIABLE.
- **"Quick reference"** lever table — a recap of the workflow doubling as a reference index; all five `references/` files are linked inline at point of use.
- **"See Also"** tail — the same links one screen lower. The dangling `deep-research` mention went with it.
- **"Project grounding" bullet 1** — a second, longer copy of the learner-grounding gate.
- Persistence mechanics inside the gate that `references/learner-grounding.md` owns in full (article template, `CLAUDE.md` snippet). The article filenames stay so the gate is still actionable without opening the reference.

## What stayed, deliberately

- **The anti-patterns table** — retitled from "Anti-patterns / rationalizations → STOP", but every row kept: these are concrete failure modes, not rules restated from above.
- **The 12-item QA gate** — it is exactly what `scripts/verify.sh` automates, and the surviving home of the non-negotiables.
- **The Expert Secrets toolkit** in full, the **gap-map flags** (they mirror the verify.sh greps), the **seven-beat landing recipe**, and the worked **Bad→Good idempotency pair** — judgement taught by demonstration.
- The learner-grounding hard gate, including the labelled `DRAFT (ungrounded …)` escape hatch.

## Verification

- `bash scripts/eval-lint.sh` → `course-storytelling  PASS (6/5/2)`
- All five `references/*.md` still linked from the body (brunson-frameworks, learner-grounding, mental-models, course-analysis, concept-landing-recipe).
- `scripts/verify.sh --path <empty dir>` exits 0; every sibling named resolves (`course-builder`, `presentations`, `design`, `marketing`, `../harness/SKILL.md`).
- `manifest.json` untouched; `npm test` / `npm run validate` left to the orchestrator (no `node_modules` in this worktree).

## Note for the orchestrator

`evals/cases.yaml` routes one `should_not_trigger` case to `deep-research`, which is not a skill in the catalog (`research-ops` / `market-research` exist). Out of scope for this format pass — flagging it since rubric dimension 6 wants a real sibling there.

### DIFF
--- a/skills/course-storytelling/SKILL.md
+++ b/skills/course-storytelling/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: course-storytelling
-description: "Use when you have course / lesson / module content and you need the teaching to LAND — so grounded, named, and emotionally resonant that the student loves the teacher. Ingests slides/transcripts/notes/outline, extracts the concepts AND the existing narrative spine, finds the abstract/jargon-heavy/ungrounded gaps, then reframes each concept with Russell Brunson's Expert Secrets methodology (Epiphany Bridge, the three false beliefs, the Big Domino, named mental models, story-selling, grounding analogies, future-pacing). Profiles the LEARNER and AUDIENCE first (hard gate). Outputs a per-concept 'landing recipe' and a reworked narrative spine. Triggers: 'make this lesson land', 'teach this concept', 'storytelling for my course', 'expert secrets', 'epiphany bridge', 'mental models for teaching', 'this lesson is too abstract', 'my students aren't getting it', 'name this concept'. NOT slide visuals (that's `presentations`/`design`) and NOT a content-audit/review pass (run review-content first, bring findings here)."
+description: "Use when lesson or course content is correct but forgettable and a concept has to LAND: profiles the learner, breaks the blocking false belief, then rebuilds it as epiphany story → named model → grounded analogy → proof → so-what. NOT outcomes, assessment or module order (that is `course-builder`), NOT slide visuals (that is `presentations`)."
 tags: [course, teaching, storytelling, content]
 recommends: [presentations]
 origin: risco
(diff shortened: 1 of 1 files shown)


ericrisco:

PR #86, opened 2026-07-25
refactor(tasks): lean the SDD task-breakdown phase to the new rubric

ericrisco:
## What

Context-engineering pass over `skills/tasks/SKILL.md` per `scripts/skill-migration-brief.md`. No content rewrite: no command, path, default, ID convention, or recommendation changed.

| | Before | After |
| --- | --- | --- |
| `description` chars | 969 | **363** |
| body bytes | 16166 | **14143** (-12.5%) |
| body lines | 290 | **266** |
| eval-lint | — | **PASS** (7 should_trigger / 6 should_not_trigger / 1 capability) |

## Description

`tasks` is in the `core` profile, so its description is in context on every turn of every install whether or not the phase fires. The old one carried a ten-phrase trigger list in English and Spanish, a full recital of the Spec Kit chain, and the artifact path.

The new one answers the only question a reader has at match time — is this the phase I want, or the one before/after it:

> Use when an approved plan needs slicing into an ordered task list before any code — the SDD `tasks` phase. Each row carries a literal done-check, dependencies, a parallel-safe marker and a spec trace, appended to the plan artifact. NOT writing the plan (that is `plan`), NOT the consistency gate (that is `analyze`), NOT executing the tasks (that is `implement`).

All three boundary siblings exist under `skills/`.

## What went

- **"Done-of-done for this phase"** — eight checkboxes, each restating a rule already stated above it (runnable done-checks, spec traces, `[P]` discipline, the closer task, the forecast, where the artifact lands). A checklist that repeats the flow gets skimmed.
- **The accompaniment dial's four-level definition** → one paragraph that reads the profile and points at `init`, which owns the gauging questions. The L0→L3 *phase-specific* mapping (table-only … confirm the done-checks) stays; the generic dial teaching does not.
- **The model-routing resolution order**, owned by `../sdd/references/model-routing.md`. The tier (`balanced`) and the opt-in gate remain; the restatement is gone so the two can't drift.
- **Prose padding** in "Next in the SDD chain" (re-explaining what `analyze` and `implement` do) and in the Interfaces rules list.

## What stayed, on purpose

- The **anti-patterns table** — required by the rubric, and these are concrete failure modes ("mark everything `[P]` so it goes faster"), not rules restated from above.
- The **six-field well-formed-row table** and the **WEAK/STRONG done-check pair** — an interface and a worked demonstration, not examples to prune.
- The **review workload forecast**, the **per-task Interfaces block** for context-isolated `developer` workers, and the **result envelope + chain position** — an SDD phase's output contract is substance, and this one already had it.

## Verify

- `bash scripts/eval-lint.sh | grep '^tasks'` → PASS
- `skills/tasks/` has no `references/` directory, so the linked-reference invariant is vacuous; the five `../<stack>/SKILL.md` routing links plus `../sdd/references/model-routing.md` and `skills/orient/references/orientation-contract.md` all resolve.
- `manifest.json` untouched, per the brief.

### DIFF
--- a/skills/tasks/SKILL.md
+++ b/skills/tasks/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: tasks
-description: "Use when you have an approved implementation plan and need to break it into an ordered, independently-verifiable task list before any code is written — the SDD `tasks` phase (GitHub Spec Kit lineage: constitution → specify → clarify → plan → tasks → analyze → implement). Turns a plan into numbered tasks, each with an explicit done-check (the literal command or observation that proves it complete), dependencies, and a parallel-safe marker. Triggers: 'break the plan into tasks', 'make a task list', 'task breakdown', 'descompón el plan en tareas', 'lista de tareas', 'what are the steps to build this', 'turn the plan into a checklist', 'sequence the work', 'which tasks can run in parallel', 'give each task a done-check'. Appends the task list into the plan artifact under 02-DOCS/wiki/sdd/plans/<slug>.md. NOT writing the plan itself (that is `plan`), NOT the consistency gate over the artifacts (that is `analyze`), NOT executing the tasks (that is `implement`)."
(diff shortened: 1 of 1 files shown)


ericrisco:

PR #88, opened 2026-07-25
refactor(sdd-init): migrate to the Claude-5-era authoring rubric

ericrisco:
Context-engineering pass on `skills/sdd-init/SKILL.md` per `scripts/skill-migration-brief.md`. No content rewrite: no recommendation, version, command, model name, figure or claim changed.

## Numbers

| Metric | Before | After |
| --- | --- | --- |
| `description` chars | 676 | **341** (target ≤350, hard limit 1024) |
| body bytes | 9672 | **9338** |
| body lines | 194 | **189** (ceiling 400) |

## Description

The old description spent ~290 chars on an 8-phrase `Triggers:` list in English and Spanish. That list is in context on **every turn the skill is installed**, invoked or not, and the model matches by meaning anyway.

Rewritten for discrimination against the two siblings this name is genuinely confused with — `sdd` (the dispatcher that routes phases) and `init` (the first-contact user/workspace bootstrap). Both verified to exist under `skills/`. `sdd-init` calibrates the SDD runtime config that the other two rely on.

## What went

- **The `Triggers:` phrase list.** Pure per-turn cost; superseded by semantic matching.
- **The accompaniment-dial restatement** in Preflight Choices (`At L0/L1 infer defaults… at L2/L3 explain the trade-off`). That dial is owned by `init` — restating its levels here is how the two drift apart. Replaced with a pointer to the profile.
- **Triple duplication of the model-routing opt-in rule.** "Ships off / never switch models behind their back" appeared in the preflight table, the models prose, the preservation rule *and* the result envelope. Kept once, next to the step it governs (the re-calibration preservation rule), and rewritten to say *why* it is absolute instead of restating the fact.
- **"Do not load every skill into context"** in Skill Registry — already carried verbatim by the anti-patterns table one screen below.

## What stayed, deliberately

- **The full `config.yaml` shape, byte for byte.** It is the runtime contract later phases read; every key is load-bearing, and the brief warns against cutting substance to hit a number.
- **The anti-patterns table** — required by the rubric, and these are concrete failure modes rather than rules restated from above. The `init` row was widened to name `sdd` too, since that is the third confusable name.
- **Preflight Choices decision table** (the flow genuinely branches), the **Detection** facts list, the **Equip the repo** install flow, and the **result envelope** — all pre-existing contracts, not padding.

## Notes

- `skills/sdd-init/` has no `references/` directory, so the every-reference-is-linked invariant is trivially satisfied. The one cross-skill link, `../sdd/references/model-routing.md`, was verified to resolve.
- `evals/cases.yaml` untouched; `manifest.json` untouched (derived, orchestrator regenerates).
- The body already met most of the rubric — no rule banks, no rationalization table, no "when to use" section, no trailing reference index — so this is a small honest diff rather than a large invented one.

### DIFF
--- a/skills/sdd-init/SKILL.md
+++ b/skills/sdd-init/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: sdd-init
-description: "Use when calibrating an existing repo before running the rsc SDD flow: detecting stack, package manager, test runners, lint/type/build commands, monorepo signals, artifact store, execution mode, review budget, strict TDD capability, and project skill registry. Triggers: 'calibrate this repo for SDD', 'run sdd init', 'detect my test runner before implementing', 'set up SDD config', 'prepare this repo for spec-driven development', 'configure per-phase model routing', 'assign models per SDD phase', 'set up cheaper model for implementation'. NOT first-contact user/workspace bootstrap (init), NOT 01-TOOLS/02-DOCS scaffolding (harness), NOT writing a feature spec (specify)."
+description: "Use when calibrating an existing repo for the rsc SDD chain: detect stack, package manager, test runners and apply/verify commands, refresh the skill registry, and write the runtime config at 02-DOCS/wiki/sdd/config.yaml. NOT the dispatcher that routes SDD phases (that is `sdd`), NOT first-contact user/workspace bootstrap (that is `init`)."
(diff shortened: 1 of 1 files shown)


ericrisco:

PR #89, opened 2026-07-25
refactor(worktrees): migrate to the Claude-5-era authoring rubric

ericrisco:
Context-engineering pass on `skills/worktrees/SKILL.md` per `scripts/skill-migration-brief.md`. No substance changed: no command, guard, path, model tier or recommendation was altered.

| Metric | Before | After |
| --- | --- | --- |
| `description` chars | 887 | **339** (target ≤350) |
| Body | 228 lines / 15624 bytes | **183 lines / 12869 bytes** |
| `eval-lint` | — | **PASS** (7 should_trigger / 5 should_not_trigger / 1 capability) |

## Description

Dropped the ten-phrase `Triggers:` list and the chain-position restatement — both are resident in context on every turn the skill is installed, invoked or not, and the model matches by meaning anyway. What is left is the rubric shape: what it does, when, and an explicit `NOT … (that is <sibling>)` boundary.

The boundary now sharpens the nearest sibling, `parallel`: **worktrees provides the isolation, `parallel` decides what may run concurrently inside it.** `ship` covers the other edge (this skill only *opens* isolation).

## What went

- **"What this skill is NOT"** — four bullets restating the description. Every route survives in the closing handoff paragraph: `tasks`, `implement`, `ship` (with the Eric-only git authorship note) and the no-`02-DOCS` rule.
- **"Red flags — stop and re-route"** — all five entries were third copies. Not-a-git-repo and the dirty-tree stop are pre-flight steps 2 and 4; the ship/implement routes live in the closing paragraph and the anti-patterns table; the destructive-remove refusal moved next to the create/keep/discard table it governs, carrying the "quoting what would be lost" detail with it. A rule in context is followed; a rule in a tail is skimmed.
- **The copy-when-isolating checklist** — every line restated a numbered pre-flight step or the contract sentence verbatim, and the numbered gate is already an ordered checklist that carries the reasoning the copy did not.
- **The standalone "Native vs. git fallback" table** — merged into the creation section, dropping the enter/exit bullets it duplicated. The table now carries both paths end to end.
- **"Where you are in the chain"** — folded into the closing handoff; the chain diagram is already at the top.

## What stayed, deliberately

- The **anti-patterns table** (rubric requires one; these are concrete failure modes, not restated rules).
- The **branch-vs-worktree decision table** — the call genuinely branches five ways.
- The **dial table** — phase-specific behaviour for `worktrees`, not the generic accompaniment dial (which belongs to `init` and is only *read* here).
- The **provenance-aware cleanup guards** and all three bash fences, verbatim.
- The **model-tier routing block** and the chain diagram.

No `references/` directory exists for this skill, so the every-reference-is-linked invariant is vacuous. `manifest.json` untouched, as instructed.

### DIFF
--- a/skills/worktrees/SKILL.md
+++ b/skills/worktrees/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: worktrees
-description: "Use when feature work needs an isolated workspace before code is touched — branch or git worktree off the default branch so the main checkout stays clean while a plan is executed. Triggers: 'isolate this work', 'I don't want to touch main', 'set up a worktree', 'work on a branch for this', 'spin up an isolated checkout', 'start the feature without dirtying my tree', 'create a worktree for the plan', 'I'm on main and need to implement', 'keep the working tree clean', 'parallel checkout for this branch'. The on-demand SDD step that runs right before implement: it confirms a clean, isolated workspace exists (native EnterWorktree first, git worktree fallback), records nothing to runtime, and hands back to implement. NOT the task breakdown (tasks), NOT writing code (implement), NOT branch cleanup/merge after the work (that's the ship phase). Honors the harness accompaniment dial."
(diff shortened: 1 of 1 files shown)


ericrisco:

PR #90, opened 2026-07-25
refactor(ship): drop the duplicate rule banks, keep every consent rule next to its step

ericrisco:
Migration of `ship` to the Claude-5-era authoring rubric (`scripts/skill-rubric.md`, per `scripts/skill-migration-brief.md`).

`ship` is in the **core** profile and is the **last** phase of the SDD chain — the one skill whose mistakes are outward and irreversible (push, PR, merge, delete). That was the lens for every cut.

## Numbers

| | before | after |
| --- | --- | --- |
| description | 769 chars | **350 chars** (-55%) |
| body | 19427 B / 280 lines | **17412 B / 250 lines** (-10%) |
| eval-lint | — | **PASS** (7 should_trigger / 5 should_not / 1 capability) |

## What went, and why

- **The trigger list** — fourteen phrases in two languages ('ship it', 'haz el merge', 'cierra la rama', …). The description is in context on every turn of every core install, invoked or not; the model matches by meaning. What replaces it is the discriminating part: close of the chain, three landing options, Eric-only authorship, and the boundary against `verify` / `review` / `deployment`.
- **"Red flags — abort and re-plan"** — five bullets, each a restatement of a checklist item or of the authorship rule stated 190 lines above it. The two facts it carried that the checklist did *not* (rotate a pushed secret; don't paper over work that landed on the trunk) moved **into** the checklist items they belong to, where they are read at the moment they apply.
- **"When to use / When NOT to use"** — the "use when" half restated the description, the "do not use" half restated the routing. All four sibling routes survive as one sentence in the intro, so no near-miss loses its destination.
- **The model-tier restatement** — the resolution algorithm it spelled out (`overrides` beats `phases`, map via `tiers`) is owned by `../sdd/references/model-routing.md` and was already drifting from it. Now: tier, opt-in flag, pointer.

## What stayed, and why

- **The authorship hard rule, unabridged.** It is the reason this phase exists. It now states *why* it is absolute — a commit is a permanent published claim about who wrote something, and a pushed trailer only comes out with a history rewrite — instead of shouting "non-negotiable".
- **Every consent-before-irreversible rule, next to the step it governs.** The discard confirmation quoting the branch name stays and now names what makes it unrecoverable. A line at the head of *Executing each option* states that nothing there runs before the user picks one, because a recommendation is not a yes.
- **The PreToolUse guard section**, kept consistent with `targets/ship-guard.mjs` — the fail-open cases are now named as the hook actually implements them (detached HEAD, no repo, git error).
- **The anti-patterns table** (8 rows; the two authorship rows merged into one), the three-option table, the delivery-strategy modes, the whole executing section, the archive bundle, and the **result envelope** — present before, kept verbatim.
- **The dial section**, rewritten as a table of what *this phase* shows at each level. That is phase-specific rendering, not a re-teach of the generic dial.

A 10% body cut is small on purpose: most of this skill is git mechanics and safety, which is the substance. No `references/` dir, so the link invariant is trivially met; all six `../<sibling>/SKILL.md` links resolve. `manifest.json` untouched.

### DIFF
--- a/skills/ship/SKILL.md
+++ b/skills/ship/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: ship
-description: "Use when the work is complete and verified and it is time to CLOSE the development branch — the final phase of the rsc SDD chain, after review approves the diff. Triggers: 'ship it', 'close the branch', 'open the PR', 'merge this', 'merge into main', 'create the pull request', 'how do I land this work', 'finish this feature', 'haz el merge', 'abre el PR', 'cierra la rama', 'súbelo a main', 'clean up the branch', 'I'm done, what now'. HARD RULE it enforces: git authorship is ALWAYS Eric — never a Co-Authored-By or 'generated with' footer in any commit or PR. NOT running lint/type/test (that is `verify`), NOT reading the diff adversarially (that is `review`), NOT deploy/release mechanics to a server (that is `deployment`). Honors the harness accompaniment dial."
+description: "Use when a verified, review-approved branch has to land — the close of the rsc SDD chain: safety checks, then three landing options (merge, PR, park/discard), authorship always Eric and never an AI trailer. NOT running the gates (that is `verify`), NOT reading the diff for defects (that is `review`), NOT shipping to a server (that is `deployment`)."
 tags: [sdd, ship, release, pr]
 recommends: []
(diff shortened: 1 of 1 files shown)


ericrisco:

PR #92, opened 2026-07-25
refactor(debug): migrate to the Claude-5-era authoring rubric

ericrisco:
Context-engineering pass over `skills/debug/SKILL.md` against `scripts/skill-rubric.md`, per `scripts/skill-migration-brief.md`. Format only — no phase semantics, commands, tiers or artifacts changed. `debug` is a `core`-profile skill with no result-envelope block; none was added.

| Metric | Before | After |
| --- | --- | --- |
| `description` chars | 908 | **349** |
| Body bytes | 15774 | **13086** |
| Body lines | 231 | **192** |
| `scripts/eval-lint.sh` | — | **PASS** (should_trigger=8, should_not_trigger=5, capability=1) |

## Description

Deleted the 11-phrase `Triggers: '…'` keyword list — the model matches by meaning, and that list was paid for on every turn `debug` is installed, invoked or not.

Added **`specify`** to the boundary set: the always-on SDD gate routes bug-fix intent to `debug` rather than to `specify`, so that is the discrimination that matters most. The `verify` (lint/test gate) and `review` (adversarial diff reading) boundaries are kept. Kept the symptom list, the "root cause reproduced and proven before any fix" rule, and "callable mid-`implement`" — those are what make the skill choosable from the description alone. Dropped the trailing "Honors the harness accompaniment dial" sentence; the body still reads the profile.

## What went (redundant, not substantive)

- **"When to use / when not to"** — five bullets restating the description; its routing survives in the chain section (implement/verify/review) and Red flags (clarify/analyze).
- **"What this skill is NOT"** — all three points already live at point of use (implement hand-back, "verify licenses done", "keep the fix minimal").
- **The rule paragraph after the loop fence** — restated each step's rule; the one rule unique to it (never skip, never reorder) moved into the section heading it governs.
- **The intro's duplicate stack-skill list** — the delegation table below names and links the same six stacks with their actual instruments.
- **The dial's trailing no-profile paragraph**, folded into its lead sentence.
- **One Red-flags bullet** ("editing code and can't reproduce") — anti-pattern row 1 word for word.
- **"Rationalization / Reality"** table framing, reworded as anti-pattern → why it fails.

## What stayed (and why)

- **All 9 anti-pattern rows** — the rubric requires the table, and each row blocks a distinct real failure (try/except swallow, sleep/retry a race, three changes at once, works-on-my-machine, skip-the-loop).
- **The five-step fence plus its four subsections** — the subsections carry mechanics the fence only names: the flake as a measured `k/N` rate, `git bisect` with the repro as oracle, one variable per isolate step, test-first red-for-the-right-reason.
- **The dial table** (what to show per level *while diagnosing* — phase-specific, not a dial tutorial), the **model-tier block**, the **stack delegation table**, the **decisions.md entry template**, the **per-bug checklist** as the exit contract, and the **chain position**.

## Verification

- `bash scripts/eval-lint.sh` → `debug PASS (8/5/1)`.
- Description 349 chars, parses as single-line quoted YAML.
- `skills/debug/` has no `references/` directory, so the linked-reference invariant is vacuous; all `../<sibling>/` links (`sdd`, `fastapi`, `go`, `nextjs`, `flutter`, `postgresdb`, `secure-coding`) resolve.
- `manifest.json` deliberately untouched; `npm test` / `npm run validate` not run (no `node_modules` in the worktree) — the orchestrator runs them globally after merge.

### DIFF
--- a/skills/debug/SKILL.md
+++ b/skills/debug/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: debug
-description: "Use when something is broken and the impulse is to start patching — a bug, a failing or flaky test, a crash, an exception, a wrong result, a regression, or behavior nobody can explain — and you want the root cause found BEFORE any fix is proposed. The on-demand rsc-sdd diagnosis discipline, callable from inside implement (or any phase) the moment a check fails for a reason you don't understand. Triggers: 'debug this', 'why is this failing', 'this test is flaky', 'find the root cause', 'it works locally but not in CI', 'this used to work', 'track down this bug', 'the fix didn't stick', 'figure out why', 'NullPointer/segfault/500 out of nowhere', 'intermittent failure'. Fixes the ONE confirmed cause, then hands back. NOT writing a planned feature (that is `implement`), NOT the lint/test gate (that is `verify`), NOT adversarial diff reading (that is `review`). Honors the harness accompaniment dial."
+description: "Use when something is broken — a failing or flaky test, crash, wrong result or regression — and the root cause must be reproduced and proven before any fix. On-demand; callable mid-`implement`. NOT a feature to spec or build (that is `specify`/`implement`), NOT the lint/test gate (that is `verify`), NOT adversarial diff reading (that is `review`)."
 tags: [debug, bug, troubleshoot]
 recommends: []
 profiles: [core, full]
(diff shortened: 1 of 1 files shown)


ericrisco:

PR #93, opened 2026-07-25
refactor(deployment): cut the trigger-phrase list and the duplicated rule bank, sharpen the DECISION/mechanics boundary

ericrisco:
Migration of `deployment` to the Claude-5-era authoring rubric (`scripts/skill-migration-brief.md`).

## Numbers

| | before | after |
| --- | --- | --- |
| description chars | 756 | 346 |
| SKILL.md bytes | 21005 | 16840 |
| SKILL.md lines | 437 | 374 |

`bash scripts/eval-lint.sh` → **PASS** (`should_trigger=7, should_not_trigger=5, capability=2`, unchanged — `evals/` untouched).

## Description

Before: 756 chars, roughly half a 12-item `Trigger phrases:` list in the model's context on every turn the skill is installed.

After (346 chars):

> Use when taking an app from source to live: choosing the deploy target from requirements (Hetzner+Coolify vs Vercel vs a third), then wiring container → CI → registry → host with build secrets, healthchecks and rollback. NOT one platform's mechanics (that is `coolify`, `vercel`, `railway`, `render`), NOT the Dockerfile alone (that is `docker`).

The boundary is the point: `coolify`, `vercel`, `railway`, `render`, `fly-io` and `hetzner` did not exist as siblings when this skill was written, and its old description read as if it owned Vercel and Hetzner outright. It owns the **decision** and the wiring between stages; they own a given platform's mechanics. All named siblings verified present under `skills/`.

## What went

- The trigger-phrase list.
- **"When to use"** — five bullets restating the description. The "When NOT to use" substance survives as a four-line "Out of scope — say so and stop" block; the k8s and Terraform refusals are behaviour the eval suite asserts, not a restatement.
- **"Core principles"** — ten numbered rules already carried by the four decision tables, the canonical Dockerfile block, and the anti-patterns table. The two things not represented elsewhere were relocated rather than dropped: "one process per container" now sits next to the Dockerfile shape it governs, and the `NET_BIND_SERVICE` exception moved into the `cap_drop: ALL` anti-pattern row.
- **Six Coolify UI bullets** (persistent storage, custom domains/SSL, per-PR previews, CPU/memory limits, blue-green) that restated `references/coolify.md` and the `coolify` sibling. Now one pointer line that still names each topic, so routing survives without the re-teaching. The six bullets that touch the pipeline itself (build pack, ports, healthcheck gating the rolling swap, Is Secret, auto-deploy XOR webhook, rollback) stayed.
- **Two of three parallel env-validation snippets.** Python, TypeScript and Go saying the same thing three times is exactly the "enumerating examples where an expressive interface would do" pattern. The per-stack recommendation (pydantic-settings / zod / caarlos0-env) and the fail-fast semantics are preserved as prose.
- **The trailing "See Also" + "References" indexes**, folded into inline links at point of use plus one "Hand off" block.

## What stayed, on purpose

- **The whole "Choosing a deploy target (3 options)" section.** `init` and `harness` both delegate the concrete deploy decision here, so the Hetzner+Coolify / Vercel / third-option trade-off, the requirements checklist, and the quick steer are this skill's reason to exist.
- All four decision tables — A (base image by stack), B (Coolify build pack), C (deploy strategy), D (secret delivery). The flow genuinely branches at each.
- The anti-patterns table (rubric-required) and the worked GOOD/BAD pairs: BuildKit `--mount=type=secret` vs `ARG`, compose bind mounts, scoped workflow `permissions`.
- The dated `trivy-action` supply-chain paragraph (GHSA-69fq-xp46-6x23 / CVE-2026-33634) — that is the sourced evidence behind the SHA-pin recommendation.
- The compose section, which `evals/cases.yaml` explicitly triggers on.
- Quick reference, pre-ship checklist, and the 02-DOCS grounding contract (compressed, not cut).

## Invariants

- All four `references/` files still linked inline from the body: `dockerfiles-by-stack.md`, `github-actions.md`, `coolify.md`, `hosting-targets.md`.
- Every `../<sibling>/SKILL.md` link resolves to an existing skill (checked programmatically).
- No recommendation, version, command, figure, or claim changed — context-engineering pass only.
- Only `skills/deployment/SKILL.md` touched; `manifest.json` and `evals/` untouched.

### DIFF
--- a/skills/deployment/SKILL.md
+++ b/skills/deployment/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: deployment
-description: "Use when containerizing an app, writing a Dockerfile, setting up GitHub Actions CI/CD, choosing a deploy target, or deploying to Coolify, Vercel, or a Hetzner VPS — multi-stage builds (FastAPI/uv, Go/distroless, Next.js standalone, Flutter web, Postgres), BuildKit build secrets, image scanning (trivy/hadolint), OIDC to registries (no long-lived secrets), least-privilege GITHUB_TOKEN, zero-downtime/rolling deploys, env/secrets flow GitHub→Coolify, healthchecks, rollback, and a hosting decision matrix (Vercel vs Hetzner+Coolify vs a third option). Trigger phrases: 'dockerize', 'write a Dockerfile', 'CI pipeline', 'GitHub Actions', 'deploy', 'ship it', 'where should I host this', 'Coolify', 'Vercel', 'Hetzner', 'VPS', 'docker-compose for local dev'."
+description: "Use when taking an app from source to live: choosing the deploy target from requirements (Hetzner+Coolify vs Vercel vs a third), then wiring container → CI → registry → host with build secrets, healthchecks and rollback. NOT one platform's mechanics (that is `coolify`, `vercel`, `railway`, `render`), NOT the Dockerfile alone (that is `docker`)."
 tags: [deploy, docker, ci, github-actions, coolify]
 recommends: [secure-coding]
 origin: risco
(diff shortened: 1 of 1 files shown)


ericrisco:

PR #94, opened 2026-07-25
refactor(parallel): discriminative description, drop three overlapping rule banks

ericrisco:
Context-engineering pass on `skills/parallel/SKILL.md` against `scripts/skill-rubric.md`. No content rewrite: no recommendation, command, path, figure or claim changed.

## Numbers

| | before | after |
| --- | --- | --- |
| description | 816 chars | **349 chars** |
| body | 19204 bytes / 234 lines | **14737 bytes / 168 lines** |

## Description

Before it carried a nine-phrase trilingual `Triggers:` list plus a sentence restating the chain position. Both are paid for on **every turn the skill is installed**, invoked or not, and the body already instructs semantic matching. The new line is shaped what-it-does · when · explicit NOT boundary, and is written to discriminate against the nearest sibling rather than to cover phrasings:

> Use when two or more pieces of work are genuinely independent — disjoint files, no shared state, no ordering edge — and should fan out across subagents, then be gathered and reconciled under a combined test suite. Decides WHAT can run concurrently; NOT the isolation it runs in (that is `worktrees`), NOT the per-unit TDD loop (that is `implement`).

The `parallel` / `worktrees` boundary is now stated the same way in three places at point of use — the description, the opening paragraph (`parallel` decides **what can run at the same time**; `worktrees` provides **the isolation it runs in**), and the chain footer.

## Removed (and why)

- **"When parallel is the wrong tool"** (5 bullets) — shared file/contract, ordering edge and can't-write-a-self-contained-brief are literally lines of the independence test above it; two-tiny-tasks and isolation-vs-concurrency are rows of the anti-patterns table below it.
- **"Red flags — stop and re-plan the partition"** (5 bullets) — every entry restates an anti-patterns row or an independence-test line. Third copy of the same rules.
- **"What this skill is NOT"** (4 bullets) — restated the description's boundary. The `implement` and `worktrees` splits moved into the opening paragraph; the `tasks` and `debug` splits into the chain footer, next to where a reader would act on them.
- **The copy-per-fan-out checklist** (10 items) — a linear restatement of the four phases immediately above it. The flow does not branch there, so it added no routing.
- **The L0–L3 accompaniment table** — the generic dial belongs to `init`; restating it here is how the two drift apart. Replaced by two lines that point at `02-DOCS/wiki/harness/user-profile.md` and keep the load-bearing rule verbatim in substance: the dial changes narration only, never the independence test or the combined-suite gate.
- **Re-teaching of `implement`'s per-task review mechanics** — compressed to the `parallel`-specific rules (that unit's diff alone, reviewer ≠ its implementer, no pre-rating severity, combined adversarial review stays `review`'s job) plus the two links, which survive.
- **The standalone "Model is REQUIRED on every dispatch" block** — folded into the `developer`-agent paragraph it repeated; every fact (omitted model inherits the session's, `developer` pins balanced, name the tier on a raw subagent) is preserved.

## Kept deliberately

- **The independence test** and the good/bad fan-out pairs — the gate the skill exists for, and judgment taught by demonstration rather than by rule.
- **The four phases**, the subagent brief template, the freeze-contracts-before-dispatch rule, and per-unit model routing (the pointer to `../sdd/references/model-routing.md`, not a restatement of it).
- **The four-status unit report contract** and the `skill_resolution` YAML — output contracts, not padding.
- **The anti-patterns table** (rubric-required, and now the single home for the failure modes the two deleted bullet lists duplicated) and the **result envelope**, which already existed — not added here.

## Verification

- `bash scripts/eval-lint.sh` → `parallel PASS (should_trigger=6, should_not_trigger=5, capability=1)`
- Frontmatter parses as YAML; `description` 349 chars (≤ 350 target, ≤ 1024 hard limit).
- No `references/` dir in this skill, so the link invariant is vacuous; every `../<sibling>/…` path referenced (`fastapi`, `implement`, `verify`, `sdd/references/model-routing.md`, `implement/references/per-task-review.md`, `implement/scripts/review-package`) resolves on disk.
- Only `skills/parallel/SKILL.md` is touched. `manifest.json` deliberately untouched — the orchestrator regenerates it.

### DIFF
--- a/skills/parallel/SKILL.md
+++ b/skills/parallel/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: parallel
-description: "Use when you have two or more genuinely independent pieces of work and want to fan them out across subagents instead of doing them one after another — disjoint scope, no shared files or state, no ordering dependency, then gather and reconcile the results. Triggers: 'do these in parallel', 'fan this out', 'split this across subagents', 'run these at the same time', 'these tasks are independent, parallelize them', 'dispatch agents for each module', 'hazlo en paralelo', 'reparte esto en subagentes', 'a la vez'. The on-demand SDD helper that implement calls when its task list has disjoint clusters; it owns the partition→dispatch→gather→reconcile discipline. NOT for sequential or data-dependent steps (serialize those), NOT the TDD loop itself (that stays inside implement), NOT git isolation (that's worktrees)."
+description: "Use when two or more pieces of work are genuinely independent — disjoint files, no shared state, no ordering edge — and should fan out across subagents, then be gathered and reconciled under a combined test suite. Decides WHAT can run concurrently; NOT the isolation it runs in (that is `worktrees`), NOT the per-unit TDD loop (that is `implement`)."
 tags: [parallel, subagents, fanout]
 recommends: []
 profiles: [core, full]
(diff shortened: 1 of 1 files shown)


ericrisco:

PR #95, opened 2026-07-25
refactor(verify): cut the trigger list and fold the when-to-use and red-flag banks into the gate

ericrisco:
Context-engineering pass on `skills/verify/SKILL.md` against `scripts/skill-rubric.md`, per `scripts/skill-migration-brief.md`. One file touched.

## Numbers

| Metric | Before | After | Δ |
|---|---|---|---|
| `description` chars | 964 | **349** | −615 (−64%) |
| SKILL.md bytes | 14737 | **13264** | −1473 (−10%) |
| SKILL.md lines | 202 | **178** | −24 |
| `eval-lint.sh` | PASS | **PASS** | 8 should_trigger / 5 should_not_trigger / 1 capability |

## Description

Before, it opened with the phase, then spent ~380 chars on a 10-phrase `Triggers:` keyword list, then restated the whole body flow, then closed on a boundary naming `implement`/`review`/`debug` but **not** `code-review`.

After:

> Use when implementation is finished and about to be called done or merged — the rsc-sdd evidence gate: runs the stack's scripts/verify.sh (lint, type, test, audit), walks every task done-check and acceptance criterion, records a dated verdict. NOT judging the diff by eye (that is `review`, spec-less `code-review`), NOT diagnosis (that is `debug`).

The boundary is rebuilt around the real mis-route — **verify vs review vs code-review**. `verify` runs machinery and reads its output; `review` reads the diff against the sdd spec/plan chain; `code-review` reads a diff with no spec chain to key off. `implement` is no longer named in the description because "Use when implementation is finished" already excludes it, and the body still routes there by name. "Honors the harness accompaniment dial" was dropped: every rsc-sdd phase does, so it carries zero routing signal.

## What was removed

- **"When to use / when not to"** (14 lines). The four "Use when" bullets restated the description; the four routing bullets survive as one line after the intro, upgraded from bare names to resolvable `../<sibling>/SKILL.md` links and now including `../code-review/SKILL.md`, which the section never mentioned.
- **"Red flags — abort the verdict"** (4 bullets). Three were verbatim duplicates of rules stated at the step they govern: RUN rule 4, the WALK acceptance bullet, anti-pattern row 7. The fourth — *you cannot point at the line of output that proves the claim* — was the only one carrying its own substance, so it survives as an anti-patterns row rather than being lost.
- The dial section's trailing paragraph, folded into its lead sentence. `init` owns the dial; this skill points at it.
- One duplicated WALK sentence ("a criterion you reviewed by reading the code is not verified"), merged into the acceptance-criteria bullet it qualifies — where the `review` boundary is actually applied.
- The FAIL-routing list in "Next in the chain", which restated the VERDICT step three lines above it, and "with Eric-only authorship" (that is `ship`'s rule to teach; re-teaching it here is how the two drift apart).
- `Rationalization | Reality` table header, reworded to `Anti-pattern | Why it fails / fix`.

## What was deliberately kept

- **All 8 anti-pattern rows, plus the rescued red flag (9 total).** The rubric requires the table, and each row prevents a distinct real failure: stale green, SKIP-as-pass, read-the-code-as-proof, lint-green-as-done, coverage below floor, partial verdict, fix-mid-gate, verdict-before-output.
- **The five gate steps in full** — SKIP-is-not-a-pass, monorepo multi-stack detection, `config.yaml` precedence over an invented command.
- **The stack gate table** — the flow genuinely branches on which stack changed, and it names the six skills that own the tooling `verify` only orchestrates.
- **The verification-record template** with its OKF frontmatter and the worked PASS/FAIL example: it teaches the output contract by demonstration.
- **The result-envelope block, the model-tier block, and the chain position.** The envelope already existed; none was added.
- The dial's L0–L3 rows (what to show at each level of a *verdict* — phase-specific, not a dial tutorial) and the invariant that the verdict itself never changes with the dial.

Phase semantics untouched: same chain position (after `implement`, before `review`/`ship`), same `balanced` tier with opt-in routing, same LOCATE → RUN → WALK → RECORD → VERDICT, same artifact path. No recommendation, path, command or figure changed.

## Verification

- `bash scripts/eval-lint.sh` → `verify PASS (should_trigger=8, should_not_trigger=5, capability=1)`
- `description` = 349 chars, parses as single-line quoted YAML, third person, `Use when` lead, explicit `NOT … (that is <sibling>)` boundary, no `Triggers:` list.
- No `references/` directory in this skill, so the link invariant is vacuous.
- All 14 `../<sibling>/SKILL.md` links resolve under `skills/`; `recommends: [review]` resolves.
- All opening code fences language-tagged.
- `manifest.json` untouched (derived; the orchestrator regenerates it).

### DIFF
--- a/skills/verify/SKILL.md
+++ b/skills/verify/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: verify
-description: "Use when implementation is finished and someone is about to say it's done, shipped, working, or ready to merge in the rsc-sdd flow — the post-implementation GATE that demands evidence before any 'done' claim. Triggers: 'verify this feature', 'is it done?', 'run the checks', 'lint/type/test before merge', 'did the acceptance criteria pass', 'confirm the tasks are complete', 'gate before review', 'audit the change', 'are we green', 'prove it works'. Runs the relevant stack skill's scripts/verify.sh (lint, type-check, tests, coverage, dependency/vuln audit), then walks every task done-check and every spec acceptance criterion, attaching the actual command output as evidence. Writes the verification record under 02-DOCS/wiki/sdd/ and points to the review phase. NOT writing tests or features (that is implement), NOT adversarial code reading (that is review), NOT root-cause diagnosis of a failing test (that is debug). Honors the harness accompaniment dial."
+description: "Use when implementation is finished and about to be called done or merged — the rsc-sdd evidence gate: runs the stack's scripts/verify.sh (lint, type, test, audit), walks every task done-check and acceptance criterion, records a dated verdict. NOT judging the diff by eye (that is `review`, spec-less `code-review`), NOT diagnosis (that is `debug`)."
 tags: [sdd, verify, test]
 recommends: [review]
 profiles: [core, full]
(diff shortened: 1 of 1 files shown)


ericrisco:

PR #98, opened 2026-07-25
refactor(postgresdb): migrate to the Claude-5-era authoring rubric

ericrisco:
Context-engineering pass on `skills/postgresdb/SKILL.md` per `scripts/skill-migration-brief.md`. No SQL, version, figure, or recommendation was changed.

## Numbers

| | Before | After |
| --- | --- | --- |
| `description` | 659 chars | **341 chars** |
| Body | 19735 bytes / 374 lines | **17670 bytes / 354 lines** |
| eval-lint | PASS | **PASS** (should_trigger=8, should_not_trigger=6, capability=2) |

Description now:

> Use when PostgreSQL engine behaviour decides the answer — schema and type design, index choice, reading EXPLAIN on a slow query, zero-downtime DDL and backfills, or ops (roles, RLS, pooling, vacuum, partitioning, PITR). PG16, ORM-agnostic. NOT portable query logic (that is `sql`), NOT a managed provider's platform surface (that is `neon`).

Both boundary siblings (`sql`, `neon`) exist under `skills/`. The old description ended in an 11-item `Triggers - "slow query", "add an index", …` list, which is paid for on every turn the skill is installed whether or not it fires.

## What went

- **"When to use" bullets** — restated the description.
- **"See Also" tail index** — all four `references/` files are already linked inline in the Deep dives paragraph; the flat sibling list routed nothing. Its final bullet also asserted that ORM-surface traps are "not skills in this collection", which is stale: `prisma-orm` and `drizzle-orm` ship in the catalog. Dropped the false clause rather than restate it; the out-of-scope reasoning survives in the compact "Not this skill" paragraph, which now also names `mysql` / `sqlite-turso` / `clickhouse-analytics` / `deployment` as the real destinations.
- **"Non-negotiables" as a rule bank** — dissolved, not deleted (see below). Four of its ten items (`timestamptz`, `numeric(19,4)`, `text`+CHECK, FK indexing) were already stated *with their reasons* in the type table's Why column and in the canonical DDL's inline comments.
- **The SDD gate blockquote** — collapsed to one sentence pointing at `../specify/SKILL.md`; the method belongs to `../sdd/SKILL.md`.
- **Project grounding** — 3 numbered steps re-teaching what `harness` owns became one paragraph with the same instruction and the same "recorded, not gated" caveat.

## What stayed (and why)

This is a high-install stack skill and the destructive-DDL rules are load-bearing, so every one of them survived — relocated next to the step it governs and given its *reason* instead of a NON-NEGOTIABLE shout:

- New **"DDL on a live table"** section carries `CONCURRENTLY`-always, the `ADD COLUMN … NOT NULL` / volatile-default table-rewrite rule, `lock_timeout` + `statement_timeout`, and forward-only migrations — each now stating which lock it takes and what that blocks. The lock-mode paragraph moved out of Quick reference to sit with them.
- **"Read the plan before adding an index"** now sits directly under the index decision table.
- The **RLS owner-bypass** rule (owner bypasses RLS; verify with a non-owner role; `FORCE ROW LEVEL SECURITY`) moved into the anti-patterns table — the only place in this body where RLS is actually read.

Untouched: both decision tables, the when-NOT-to-index list, all eight GOOD/BAD SQL pairs, the EXPLAIN four-point method, the anti-patterns table (retitled from `## Anti-patterns / rationalizations → STOP` to the catalog-standard `## Anti-patterns` — 185 skills already use that heading — rows verbatim plus the relocated RLS row), isolation levels, the six diagnostic queries, and the `verify.sh` contract.

## Verification

- `bash scripts/eval-lint.sh | grep '^postgresdb'` → **PASS**
- All 4 `references/*.md` files linked from the body (schema-and-indexing, query-optimization, migrations ×3 incl. the `#lock-impact-reference` anchor, operations-and-security).
- All 7 `../<sibling>/SKILL.md` links resolve to real skills.
- `description` is 341 chars, parses as single-line YAML.
- `evals/`, `scripts/verify.sh`, `manifest.json` untouched; only `skills/postgresdb/SKILL.md` is in the diff.

### DIFF
--- a/skills/postgresdb/SKILL.md
+++ b/skills/postgresdb/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: postgresdb
-description: Use when designing or reviewing a PostgreSQL schema (types, constraints, normalization), writing or optimizing SQL, choosing or adding indexes, reading EXPLAIN/ANALYZE, planning a migration (expand-contract, concurrent index, batched backfill), or operating/securing Postgres (roles, RLS, PgBouncer pooling, VACUUM/autovacuum, pg_stat_statements, declarative partitioning, backups/PITR). Triggers - "slow query", "add an index", "EXPLAIN", "design this table", "zero-downtime migration", "row level security", "connection pool", "N+1", "JSONB", "full-text search", "pgvector". PostgreSQL 16; ORM-agnostic (SQLAlchemy/Alembic, Prisma, golang-migrate, raw SQL).
+description: Use when PostgreSQL engine behaviour decides the answer — schema and type design, index choice, reading EXPLAIN on a slow query, zero-downtime DDL and backfills, or ops (roles, RLS, pooling, vacuum, partitioning, PITR). PG16, ORM-agnostic. NOT portable query logic (that is `sql`), NOT a managed provider's platform surface (that is `neon`).
 tags: [postgres, sql, database, migrations]
 recommends: [secure-coding]
 origin: risco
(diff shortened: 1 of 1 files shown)


ericrisco:

PR #99, opened 2026-07-25
refactor(typescript): Claude-5-era rubric pass — description 858→346, body 13.8→11.1 KB

ericrisco:
## Metrics

| | before | after |
| --- | --- | --- |
| `description` chars | 858 | **346** (target ≤350, hard limit 1024) |
| SKILL.md bytes | 13783 | **11065** |
| SKILL.md lines | 258 | **225** (ceiling 400) |

## What went

- **The `Triggers:` list** — eight quoted phrases in English and Spanish, 512 of the
  description's 858 chars. It is in context on every turn the skill is installed, and the
  model matches by meaning without it.
- **`## When to use / When NOT to use`** (23 lines). The "Use when" bullets restated the
  description; the "When NOT to use" delegate list duplicated the boundary section directly
  beneath it. The routing that only existed there (`api-design`, `testing-web`/`e2e-testing`,
  `secure-coding`, `python`, `go`, `prisma-orm`) was folded into the surviving
  **Boundary: language vs runtime vs framework** section, so no near-miss lost its target.
- **7 of the 10 `## Decision rules`.** Each was stated three times: as a rule, as a worked
  bad/good pair in the toolbox, and as a row in the anti-patterns table. Where a rule's
  *why* was not already demonstrated, it moved next to what it governs instead of being
  dropped — the `any`-poisoning sentence now heads the `unknown` section, and the
  `{ data?; error?; loading? }` illegal-state argument now heads the discriminated-unions
  section. The three rules with no demonstration below them survive (prefer inference /
  annotate boundaries; `import type` + `verbatimModuleSyntax`; narrow-don't-cast, merged
  with the `as`-last-resort rule that continued it).

## What stayed, and why

- **The anti-patterns table, intact** — 10 concrete failure modes with a why-it-bites and a
  do-instead. Required by the rubric and a different animal from a rationalization bank.
- **Every worked bad/good pair** (generics/`const` type params, `unknown`, `satisfies`,
  discriminated unions, brands, mapped/conditional one-liners) — these teach judgment by
  demonstration and are what the trimmed rules now defer to.
- **All substance verbatim**: the TS 5.9 / TS 7 `tsgo` freshness paragraph, the strict
  `tsconfig` base block and every flag comment, the `bundler` vs `nodenext` decision table,
  the project-references and declaration-emit guidance. No flag, version, figure, or
  recommendation was reworded. One correctness fix: `loading?: bool` → `loading?: boolean`.
- **Both `references/` links** — `type-system.md` (once) and `build-and-monorepo.md` (twice)
  are still pointed to inline at point of use, as is `../harness/SKILL.md`.

## Boundary check

New description names three real siblings: `nodejs` (runtime, fs/streams/http),
`react`/`nextjs` (component and route typing), `drizzle-orm` (ORM type-gen). All five
`route_to` targets in `evals/cases.yaml` (`nodejs`, `react`, `drizzle-orm`, `python`,
`testing-web`) exist under `skills/` and are still routed to from the description or the
boundary section.

`manifest.json` untouched; only `skills/typescript/SKILL.md` is in the diff.

### DIFF
--- a/skills/typescript/SKILL.md
+++ b/skills/typescript/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: typescript
-description: "Use when writing, reviewing, or fixing TypeScript type code (.ts/.tsx/.d.ts), modeling data with generics/unions/branded types, or configuring tsconfig and the build toolchain - the language and type system itself, not any runtime or framework. Triggers: \"make this generic preserve the literal type\", \"why won't this narrow / why won't TypeScript narrow this union\", \"type X is not assignable to type Y, fix the type not cast it\", \"set up a strict tsconfig with project references\", \"convert these magic strings into a discriminated union with exhaustiveness\", \"bundler vs nodenext moduleResolution\", \"configura el tsconfig en modo estricto y elimina los any\", \"por qué no estrecha el tipo\". NOT the Node runtime, fs/streams/http server (that is nodejs); NOT React/Next component wiring (that is react/nextjs); NOT ORM schema type-gen (that is drizzle-orm)."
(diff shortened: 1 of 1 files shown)


ericrisco:

PR #100, opened 2026-07-25
refactor(fastapi): migrate to the Claude-5-era authoring rubric

ericrisco:
Migrates `fastapi` to the Claude-5-era authoring rubric (`scripts/skill-rubric.md`, applied per `scripts/skill-migration-brief.md`).

| | Before | After |
|---|---|---|
| `description` | 544 chars | **386 chars** |
| body | 22,172 bytes / 501 lines | **20,056 bytes / 469 lines** |
| eval-lint | PASS | **PASS** (should_trigger=7, should_not_trigger=6, capability=2) |

## Description

Dropped the `Triggers:` enumeration — endpoint/router keywords, library names, "any `.py` file importing fastapi, pydantic, sqlalchemy, starlette". The model matches by meaning, and that list was in context on every turn the skill is installed, invoked or not.

The new description carries three boundaries against **real** siblings, chosen for discrimination rather than coverage:

- `python` — language-level Python, typing, packaging
- `postgresdb` — engine-level SQL and indexing
- `api-design` — framework-agnostic REST contracts (mirrors api-design's own `NOT implementing endpoints in a framework (that is fastapi)`)

## What went

- **"When to use" (6 bullets) + "When NOT to use" (8 bullets)** → one Out-of-scope line. Both restated the description; the line keeps every routing target and now *links* them (`django`, `python`, `api-design`, `postgresdb`, `secure-coding`, `deployment`).
- **"Quick reference" idiom table (14 lines)** — every row was either shown verbatim in a code block above it or delegated to a linked reference. The idioms that appeared nowhere else in the body (`db.get` / `scalar_one_or_none`, `selectinload` vs `joinedload`, the `argon2.PasswordHasher().hash()` and `jwt.decode(...)` call shapes) moved next to the SQLAlchemy and Security sections that govern them — a rule in context is followed, a rule in a tail index is skimmed.
- **Tail "See Also" index** — all four `references/` files are now real markdown links inline at point of use; the sibling links live in Out-of-scope, Security and Production. Nothing lost a link, it just stopped being listed twice.
- **"Project grounding" 16 lines → 6** — same instruction (read `wiki/stack/fastapi.md` first, record the real choices, index it, never gate), without re-teaching what `harness` owns.
- The cross-reference `"The non-negotiables also appear in the anti-patterns table below"`.

## What stayed, on purpose

- **The SDD new-feature gate**, verbatim. Routing, not a rule bank — it is the only thing between this skill and writing feature code with no approved spec.
- **The anti-patterns table** (the rubric requires one), reframed from quoted rationalizations to the anti-pattern itself. Every "reality" cell is unchanged.
- **All eight decision rules, the project layout, and all seven code blocks.** 301 of the remaining 470 lines are inside code fences — factory/lifespan, settings, Pydantic v2 split, DI, routers, error envelope, async ORM.
- **Pinned stack versions**, the Starlette `GHSA-86qp-5c8j-p5mr` advisory, and every API name — untouched. Context-engineering pass, not a content rewrite.
- **`references/testing.md` and its inline link** — `implement` and `debug` both delegate their Python test mechanics to that exact path.

## Note

The body is **469 lines, over the rubric's 400-line ceiling.** Getting under it would mean deleting ~70 lines of working FastAPI code, which is substance, not padding. Flagging rather than cutting.

`manifest.json` untouched (derived; orchestrator regenerates). `npm test` / `validate` not run — no `node_modules` in this worktree, per the brief.

### DIFF
--- a/skills/fastapi/SKILL.md
+++ b/skills/fastapi/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: fastapi
-description: "Use when building, reviewing, testing, securing or deploying a FastAPI / async Python service. Triggers: creating endpoints/routers, Pydantic v2 models (Create/Update/Response), dependency injection, async SQLAlchemy 2.0 + Alembic, OAuth2/JWT auth + RBAC, pytest + httpx ASGITransport tests, CORS / rate limiting / secret handling, uvicorn/gunicorn + structured logging + healthchecks + graceful shutdown, pyproject + ruff + mypy strict + uv. Any .py file importing fastapi, pydantic, sqlalchemy, starlette, or a pyproject.toml declaring those."
+description: "Use when building, reviewing, testing, securing or shipping a FastAPI / async Python service — routers, Pydantic v2 schemas, dependency injection, async SQLAlchemy 2.0, OAuth2/JWT, ASGITransport tests, production wiring. NOT language-level Python or packaging (that is `python`), NOT engine-level SQL (that is `postgresdb`), NOT framework-agnostic REST contracts (that is `api-design`)."
 tags: [python, api, async, backend, rest]
 recommends: [postgresdb, secure-coding, deployment]
 origin: risco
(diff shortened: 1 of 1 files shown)


ericrisco:

PR #101, opened 2026-07-26
refactor(nextjs): cut the duplicated lookup layer, sharpen the boundary

ericrisco:
Migrates `nextjs` to the Claude-5-era authoring rubric. Context-engineering pass only — no
recommendation, version, command or figure changed.

| Metric | Before | After |
| --- | --- | --- |
| `description` | 529 chars | **345 chars** |
| Body | 25090 B / 474 lines | **20702 B / 400 lines** (-17%) |
| eval-lint | PASS | **PASS** (7 should_trigger / 6 should_not_trigger / 2 capability) |
| `references/` linked from body | 5 of 6 | **6 of 6** |

### Description

The old one spent half its budget on a `Triggers on …` keyword list (`app/` directory,
`next.config.ts`, `proxy.ts`/`middleware.ts`, `use cache`, `useActionState`, hydration errors,
Core Web Vitals) that the model matches by meaning anyway — and this is the most-installed stack
skill, so that list is paid for on every turn by a large share of users. Rewritten for
discrimination, not coverage, with an explicit boundary: NOT framework-agnostic React or a Vite SPA
(that is `react`), NOT visual/UI design (that is `design`).

### What went

- **"When to use / When NOT to use"** — the six "Use when" bullets restated the description. The
  redirect list survives as a four-line *Not this skill* paragraph that now names resolvable
  siblings (`../react/`, `../react-native/`, `../fastapi/`, `../go/`, `../postgresdb/`,
  `../secure-coding/`) where it used to say "a generic React skill" in prose.
- **The 18-row "Quick reference" table** — a pure lookup layer over content already taught above
  it. The only two facts that lived *only* there, `revalidatePath` and `import 'server-only'`, were
  folded into the caching block and the security checklist.
- **The trailing "## References" index and "## See Also"** — five of six references were already
  linked at point of use, and every See Also target survives inline (postgresdb + secure-coding in
  the boundary paragraph, marketing in Metadata & SEO with its "we emit the tags, they pick the
  content" split, harness in Project grounding). `references/testing.md` was index-only; it now has
  an inline home in Verify, so the invariant holds — **all six references are linked**.
- **The `error.tsx` boilerplate and the GET twin of the route handler** — each re-demonstrated a
  pattern the file-role table and the POST example already show. Preserved as one prose line each
  (digest + `reset()` + `role="alert"`; uncached GET on v15 + `auth()` in every handler).
- **Project grounding** compressed 17 → 8 lines, same three instructions, no re-teaching of what
  `harness` owns.

### What stayed, deliberately

The v15/v16 detection gate and its signal table (the skill's single most load-bearing branch), the
four Server/Client boundary laws, both caching-model code blocks including the
`use cache` + `cookies()` Bad/Good pair, the `params`-is-a-Promise Bad/Good pair, the Server Action
example with auth + zod + ownership check, the security checklist, the perf lever list, the SDD
gate, and the anti-patterns table the rubric requires — header renamed from "Rationalization" to
"Common belief" so it reads as the ten failure modes it actually lists.

### Substance untouched

Next.js 15 vs 16 model split, `cacheLife`/`cacheTag`/`updateTag` stability, `next lint` removed in
v16, `proxy.ts` correct on v16, LCP < 2.5s / CLS < 0.1 / INP < 200ms, the 50k sitemap cap and the
1200x630 OG figure are all verbatim. `manifest.json` untouched — the orchestrator regenerates it.

### DIFF
--- a/skills/nextjs/SKILL.md
+++ b/skills/nextjs/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: nextjs
-description: "Use when building, reviewing, testing, securing, or optimizing a Next.js App Router app (Next.js 15/16, React Server Components, \"use client\"/\"use server\", server actions, route handlers, layouts, streaming/Suspense, the App Router caching model, TypeScript, Auth.js, Vitest/Playwright, CSP/security headers). Triggers on `app/` directory work, `next.config.ts`, `proxy.ts`/`middleware.ts`, server actions, route handlers, `use cache`, `useActionState`, RSC boundary questions, hydration errors, and Next.js perf/Core Web Vitals."
+description: "Use when building, reviewing, testing, securing, or optimizing a Next.js App Router app: Server vs Client boundaries, `use server` actions, route handlers, the v15 vs v16 `use cache` caching model, metadata/SEO, auth, and Core Web Vitals. NOT framework-agnostic React or a Vite SPA (that is `react`), and NOT visual/UI design (that is `design`)."
 tags: [nextjs, react, frontend, web, seo, ssr]
 recommends: [design, secure-coding, deployment]
 origin: risco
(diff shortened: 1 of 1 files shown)


ericrisco:

PR #102, opened 2026-07-26
refactor(go): cut duplicated rule banks and tail indexes, sharpen the description

ericrisco:
Context-engineering pass on `skills/go/SKILL.md` per `scripts/skill-migration-brief.md` and `scripts/skill-rubric.md`. No substance changed: every version, command, figure and code block is byte-identical or moved verbatim.

## Numbers

| | Before | After |
|---|---|---|
| `description` chars | 658 | **344** (target ≤350) |
| SKILL.md bytes | 18750 | **15511** |
| SKILL.md lines | 438 | **390** (ceiling 400) |
| eval-lint | PASS | **PASS** (`should_trigger=7, should_not_trigger=6, capability=2`) |

## Description

Dropped the `Triggers - "write a Go service/handler", ...` keyword list and the inline idiom enumeration. That list is in context on every turn the skill is installed, invoked or not, and the model matches by meaning anyway. Added an explicit NOT boundary naming two real siblings so the description discriminates instead of merely covering:

> Use when writing, reviewing, testing, or shipping Go code and HTTP services: idioms, `%w` error wrapping, goroutine/context/errgroup concurrency, net/http 1.22 routing, log/slog, project layout, table-driven tests, Go hardening. NOT language-agnostic threat modeling (that is `secure-coding`), NOT Dockerfile/CI shipping (that is `deployment`).

## What went

- **"Use when" bullets** — a verbatim restatement of the description.
- **The 10-item "Decision rules" bank.** Nine of ten were already taught in their own section further down. Rather than delete them, each moved next to the step it governs: receivers → Idioms, ctx-first / `context.TODO()` → Concurrency, wrap every crossed boundary → Errors, validate at the boundary → Security, `go vet`/`staticcheck` as build failures → Testing, "clear over clever" → Early return. A rule in an appendix is skimmed; a rule in context is followed.
- **Six Quick-reference rows** (wrap error, classify error, route, log, shutdown, leak guard) — each duplicated a code block on the same page. The seven toolchain commands stay, retitled "Toolchain gate".
- **The Errors section's three-line anti-pattern note** — all three already in the anti-patterns table.
- **"See Also" + the local-references tail index.** Both restated links that already appear at point of use. The three sibling links survive in the new `## Boundary` section; the three `references/` links survive inline.
- **Project grounding** compressed from a 3-step procedure to one paragraph — same article path, same content list, same "recorded, not gated / never block" rule.

## What stayed, deliberately

- **The anti-patterns table, all twelve rows.** The rubric requires one, and these are concrete Go failure modes. With the decision-rule bank gone this table is now the single place they appear. Retitled from "Anti-patterns / rationalizations -> STOP" and recolumned `Anti-pattern | Do instead`, which drops the borrowed rationalization-table framing (rubric dim. 7) while preserving every recommendation verbatim.
- **The SDD new-feature gate**, verbatim — a shared house block across eight stack skills, and it points rather than restates.
- **All three `references/` links, inline.** `references/testing.md` is load-bearing beyond this skill: `skills/implement/SKILL.md:94` delegates Go test mechanics to `../go/references/testing.md`.
- **Every load-bearing fact untouched** — Go 1.22+ / 1.26 stable, `pgx v5`, TLS 1.2 floor, 1 MiB body cap, `errgroup` `SetLimit(8)`, all four `http.Server` timeouts, `govulncheck`, `ldflags` stamping, the 3-layer error boundary.

## Gates

- `bash scripts/eval-lint.sh` → `go PASS`
- Frontmatter parses as YAML; `name: go` matches the dir; `origin: risco` present.
- All `recommends` (`postgresdb`, `secure-coding`, `deployment`) resolve; all five `../<sibling>/SKILL.md` links resolve.
- No unlinked file under `skills/go/references/`.

### DIFF
--- a/skills/go/SKILL.md
+++ b/skills/go/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: go
-description: "Use when writing, reviewing, testing, securing, or shipping Go code or HTTP services - Go idioms (simplicity, useful zero value, accept-interfaces/return-structs, functional options, embedding), error wrapping with %w and errors.Is/As, goroutines/channels/context/errgroup concurrency, net/http with Go 1.22 routing, log/slog structured logging, graceful shutdown, cmd/-internal/-pkg/ layout, table-driven tests with httptest/-race/fuzz, and govulncheck/SQL-parametrization security. Triggers - \"write a Go service/handler\", \"review this Go code\", \"add tests\", \"fix a goroutine leak\", \"is this idiomatic Go\", .go files, go.mod, net/http, slog, chi, errgroup."
+description: "Use when writing, reviewing, testing, or shipping Go code and HTTP services: idioms, `%w` error wrapping, goroutine/context/errgroup concurrency, net/http 1.22 routing, log/slog, project layout, table-driven tests, Go hardening. NOT language-agnostic threat modeling (that is `secure-coding`), NOT Dockerfile/CI shipping (that is `deployment`)."
 tags: [go, golang, http, backend, service]
 recommends: [postgresdb, secure-coding, deployment]
 origin: risco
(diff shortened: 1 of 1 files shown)


ericrisco:

PR #103, opened 2026-07-26
refactor(python): cut the trigger list and the rules the body already teaches

ericrisco:
Migrates `python` to the Claude-5-era authoring rubric. Context-engineering pass only — no
recommendation, version, command, flag or figure changed.

| Metric | Before | After |
| --- | --- | --- |
| `description` | 757 chars | **331 chars** (-56%) |
| Body | 18442 B / 369 lines | **15215 B / 320 lines** (-17.5%) |
| eval-lint | PASS | **PASS** (6 should_trigger / 5 should_not_trigger / 1 capability) |
| `references/` linked from body | 3 of 3 (bare paths) | **3 of 3** (inline links) |

### Description

A third of the old budget was a `Triggers: '…'` phrase list in two languages — "set up a Python
project with uv", "is this Pythonic / modernize to 3.12+", "convertir este gather a TaskGroup",
"any .py file or pyproject.toml" — that the model matches by meaning anyway. This is a high-install
language skill, so that list was paid for on every turn by a large share of users. Rewritten for
discrimination rather than coverage: what the skill owns, plus the two boundaries that actually
decide routing — NOT a FastAPI/ASGI service (that is `fastapi`), NOT a deep pytest suite (that is
`testing-py`).

### What went

- **The "When to use / When NOT to use" section** — its "Use when" paragraph restated the
  description clause for clause. The delegation list survives whole as *Scope — what this skill
  delegates*; every redirect is unchanged (fastapi, testing-py, secure-coding, deployment,
  go/typescript/rust, django).
- **Five of the ten decision rules** — each restated something the body teaches in full below it.
  They moved next to what they govern instead of vanishing: pathlib-not-`os.path` and
  logging-not-`print` are now bullets in Stdlib idioms; "never hand-edit `requirements.txt`,
  `uv add` updates `uv.lock`" sits in Project layout & packaging; TaskGroup-over-`gather` is the
  Async section's whole thesis (the old rule literally pointed at it — "rule 8 in Async below");
  the mutable-default rule is row one of the anti-patterns table.
- **The 15-row "Quick reference" table** — a lookup copy of commands and idioms already shown
  above it (uv verbs block, quality gate, typing examples, security block). Checked row by row
  before cutting; nothing lived only there.
- **The trailing "See Also" and reference index** — pointer layers over links that already exist
  at point of use. `fastapi`, `secure-coding`, `deployment` and `harness` are linked inline,
  `testing-py` is named twice by id, and the three `references/` files became real inline markdown
  links instead of bare paths, so the invariant holds and they are now clickable.
- **Project grounding** compressed 12 → 6 lines: same three instructions, minus the
  re-explanation of the 02-DOCS / `CLAUDE.md` layout that `harness` owns.

### What stayed, deliberately

The version and tooling paragraph (3.12 floor / 3.14 current released 7 Oct 2025, PEP 649/749,
PEP 750, `compression.zstd`, uv 0.11, ruff 0.15, mypy 1.20 `--strict`, pytest 8), the data-modeling
decision table, the TypeVar-vs-PEP-695, `gather`-vs-`TaskGroup` and eval/pickle/`shell=True`/random
Bad/Good pairs, the `pyproject.toml` and uv-verbs blocks, the quality gate, the pytest baseline, the
embedded Python-specific security block, and the 12-row anti-patterns table the rubric requires —
its column header renamed from "Rationalization" to "Tempting move" so it reads as the concrete
failure modes it actually lists.

### Substance untouched

Every version, PEP number, flag and command is verbatim. All ten fences stay language-tagged, the
frontmatter parses as YAML, all four `../<sibling>/SKILL.md` links resolve, and `manifest.json` is
untouched — the orchestrator regenerates it.

### DIFF
--- a/skills/python/SKILL.md
+++ b/skills/python/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: python
-description: "Use when writing, reviewing, modernizing, typing, or packaging Python at the language level (any framework or none) - PEP 695 generics and the type-alias statement, mypy --strict typing, dataclasses/Protocol/TypedDict/Enum choices, asyncio.TaskGroup and structured concurrency, stdlib idioms (pathlib/itertools/functools/contextlib/match), src/ layout + pyproject.toml managed with uv, and a ruff+mypy+pytest verify gate. Triggers: 'set up a Python project with uv', 'is this Pythonic / modernize to 3.12+', 'make mypy --strict pass', 'convertir este gather a TaskGroup', 'dataclass slots frozen vs NamedTuple', any .py file or pyproject.toml. NOT building a FastAPI/ASGI service (that is fastapi), NOT a deep pytest fixture/mock suite (that is testing-py)."
+description: "Use when the task is Python itself, in any framework or none: PEP 695 generics, mypy --strict typing, dataclass/Protocol/TypedDict/Enum choices, asyncio.TaskGroup, stdlib idioms, src/ layout + pyproject.toml with uv, ruff+mypy+pytest gate. NOT a FastAPI/ASGI service (that is fastapi), NOT a deep pytest suite (that is testing-py)."
 tags: [python, typing, async, packaging, uv]
 recommends: [fastapi, secure-coding, deployment]
 origin: risco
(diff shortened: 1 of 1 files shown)


ericrisco:

PR #104, opened 2026-07-26
refactor(secure-coding): migrate to the Claude-5-era authoring rubric

ericrisco:
Migrates `secure-coding` to the Claude-5-era authoring rubric (`scripts/skill-rubric.md`, applied per `scripts/skill-migration-brief.md`).

| | Before | After |
|---|---|---|
| `description` | 547 chars | **353 chars** |
| body | 20,366 bytes / 365 lines | **17,936 bytes / 336 lines** |
| eval-lint | PASS | **PASS** (should_trigger=7, should_not_trigger=6, capability=2) |

**Nothing security-substantive changed.** The OWASP Top 10 mapping, the Argon2id parameters and the OWASP floor they are measured against, the full SSRF blocklist (`169.254.169.254`, `10/8`, `172.16/12`, `192.168/16`, `127/8`, `::1`, `fc00::/7`, `fe80::/10`), the pinned-algorithm JWT decode, the magic-byte upload sniff, the CSP/HSTS header set, every tool flag and every pinned version are all still in the body, verbatim. A security rule is load-bearing by definition — this pass moved rules, it did not soften or drop any.

## Description

Dropped the 8-phrase `Triggers: '…'` list ("security review", "threat model this", "is this safe", "harden auth", "rotate secrets", "fix this vuln", "OWASP", …). The model matches by meaning, and that list is in context on every turn the skill is installed, invoked or not.

The new boundary names two **real** siblings, chosen for discrimination:

- `security-scan` — running the scanners and triaging their raw output. This mirrors security-scan's own description, which already ends `NOT threat-modeling, OWASP design reasoning, or authoring the fix (that is secure-coding)`; the pair now discriminates from both sides.
- `agent-safety` — prompt-injection guardrails on a running agent. "Prompt injection" reads like a security bug, so this is the mis-route most worth pre-empting.

`security-reviewing a diff` and the four stacks stayed: they are the strongest routing signal the skill has.

## What went

- **"When to use / When NOT to use" (18 lines)** → one paragraph after the operating posture. The Use-when bullets restated the description; all four NOT bullets survive with every routing target upgraded from a bare name to a resolvable link (`agent-safety`, `building-agents`, `security-scan`, `deployment`), and the legal out-of-scope line on third-party pentesting is untouched.
- **"Core principles (non-negotiable)" (8 numbered rules)** → each moved to the step it governs, with the reason it is absolute instead of a shouted header. Validate-at-the-boundary opens Input validation; parameterize is row 1 of the trust-boundary table; authorize per-object on every request is now the flagship A01 paragraph *and says why* — the id in the URL is attacker-controlled, so the ownership predicate is the only thing between a stranger and the row; encode-by-context sits in the XSS paragraph; secrets-never-in-repo opens Secrets; fail-closed is in the logging list with its reason (a stack trace in a response is free reconnaissance); reachable-CVE-is-a-release-blocker is workflow step 6; least privilege moved *up* into the 30-second model, where it is the one control that still caps blast radius after another defense fails.
- **"Quick reference" (14-row tail table)** — every row already appeared, or now appears, at point of use: the semgrep `--config=auto --severity ERROR` gate in the verify.sh section, the `gitleaks dir . --redact` / `gitleaks git .` split and the `pip-audit` / `osv-scanner scan source -L pnpm-lock.yaml` / `govulncheck ./...` flags in Secrets & supply chain, Argon2id parameters and cookie flags in AuthN/AuthZ, the CSP starter and CORS rule inside their code blocks, and the SSRF blocklist inside the anti-patterns row that argues for it.
- **Tail "See Also" index** — folded into inline mentions. Nothing lost a pointer; the `harness` link in particular moved onto the secrets bullet, where `01-TOOLS/<PROVIDER>/.env` is the rule being reinforced.
- **The repeated `references/owasp-by-stack.md A0x` cell** in every row of two tables → one line stating that the A-numbers index that file.
- **"Project grounding" 17 lines → 9** — same instruction (read `wiki/stack/security.md` first, record the real choices, index it, bump `Updated`, never gate the task), without re-teaching what `harness` owns.

## What stayed, on purpose

- **The anti-patterns table, all 12 rows**, retitled from "Anti-patterns / rationalizations → STOP". The rubric requires the table, and no row restates a rule above it — each carries a concrete attack: XSS steals a `localStorage` token instantly, internal fetch *is* the SSRF target (metadata, RDS), a transitive CVE still ships in your bundle.
- **All three IDOR vulnerable→fixed blocks** (FastAPI + SQLAlchemy 2.0, Next.js 15 App Router with `params` as a Promise, Go 1.22 `ServeMux` + `slog`) and every other fence. They teach judgment by demonstration and carry the skill's actionability.
- The read-only / exploitability-first / every-finding-ships-a-fix posture, the 6-step PR-sized workflow, the trust-boundary table, the OWASP fastest-fix table, the lethal-trifecta framing, and `verify.sh` as the gate.
- The 2026 Auth.js / Better Auth library note and every pinned version (Python 3.12+, Next.js 15 / React 19, Go 1.22+, Dart 3, PostgreSQL 16).

## Verification

- `bash scripts/eval-lint.sh` → `secure-coding PASS (should_trigger=7, should_not_trigger=6, capability=2)`
- `description` 353 chars, parses as single-line quoted YAML.
- All four `references/` files are pointed at from the body: `threat-modeling.md` (workflow step 2), `owasp-by-stack.md` (step 3, the A-number note, the flagship tail), `authn-authz.md` (trust-boundary table + AuthN tail), `secrets-and-supply-chain.md` (Secrets tail + verify.sh).
- All 10 `../<sibling>/SKILL.md` links resolve under `skills/`.
- `manifest.json` untouched (derived; the orchestrator regenerates it). `npm test` / `npm run validate` not run — no `node_modules` in this worktree, per the brief.

### DIFF
--- a/skills/secure-coding/SKILL.md
+++ b/skills/secure-coding/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: secure-coding
-description: "Use when threat-modeling a feature, reviewing code or a diff for security, hardening authentication/authorization, handling secrets, configuring CORS/CSP/security headers, or fixing OWASP-class vulnerabilities (broken access control, injection, SSRF, auth failures, supply-chain) in FastAPI/Python, Go, Next.js, or Flutter. Triggers: 'security review', 'threat model this', 'is this safe', 'harden auth', 'rotate secrets', 'fix this vuln', 'OWASP', 'why is this endpoint exposed', before merging an endpoint that touches auth/payments/PII/uploads."
+description: "Use when securing application code: threat-modeling a feature, security-reviewing a diff, hardening auth/authz, secrets, CORS/CSP, or fixing OWASP-class bugs (access control, injection, SSRF, supply chain) in FastAPI, Next.js, Go or Flutter. NOT running the scanners (that is security-scan), NOT agent prompt-injection guardrails (that is agent-safety)."
 tags: [security, owasp, stride, auth, review]
 recommends: [deployment]
 origin: risco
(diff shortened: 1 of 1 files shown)


ericrisco:

PR #105, opened 2026-07-26
refactor(flutter): cut the trigger list and the two trailing reference indexes

ericrisco:
Context-engineering pass on `flutter` against `scripts/skill-rubric.md`. No recommendation, version, command or code snippet changed.

## Numbers

| | before | after |
|---|---|---|
| description | 403 chars | **349 chars** |
| body | 20586 B / 449 lines | **17527 B / 400 lines** (-15%) |
| eval-lint | PASS | **PASS** (7 should_trigger / 6 should_not_trigger / 2 capability) |

## Description

The old one ended in a 200-char `Triggers:` keyword list. That list is in context on every turn the skill is installed, invoked or not, and the model matches by meaning anyway. It is now replaced by a discriminating boundary naming the two skills a reader is genuinely choosing between:

> …Targets Flutter 3.44 / Dart 3.12. NOT React Native (that is `react-native`), NOT Compose Multiplatform (that is `compose-multiplatform`).

Both siblings exist under `skills/`.

## What went

- **"Use when"** — three bullets restating the description one screen below it.
- **"Quick reference"** (15 rows) — every row pointed at a reference already linked inline at the point where that API is taught, and its Default-API column duplicated "Decision rules".
- **"See Also"** — a second index of the same five references.
- **Four anti-pattern rows** that were verbatim restatements of a line above them: the two-boolean-flags row and the global-provider-for-a-checkbox row are literally rows of "Decision rules"; legacy `StateProvider` and `pumpAndSettle` are each stated in full in their own section.
- **"Project grounding"** compressed from a numbered re-teaching of the 02-DOCS wiki convention to one paragraph pointing at `harness`, which owns it.

## What stayed, and why

- **Every code block and every pinned version** — Flutter 3.44 / Dart 3.12, Riverpod 3.0, go_router 17.2.x (+ `go_router_builder` 4.3.x), freezed 3.x, dio 5.x, mocktail 1.x. This is the substance the skill is installed for.
- **The SDD new-feature gate** and **"Decision rules"** — a real branch in the flow, not a rule bank.
- **The anti-patterns table**, required by the rubric — retitled from `Rationalization | Reality` to `Anti-pattern | Why it fails / do instead`, so it names the failure mode rather than a voice in the reader's head. Ten concrete rows remain.
- **All five `references/` files linked inline**; `references/testing.md` is pointed to from both the Riverpod and the Testing section, since widget/golden test mechanics live there.
- **`scripts/verify.sh`** lost its index row and gained a production-checklist line, so it stays discoverable where it is actually used.

The "Do NOT use when" bullets became a "Boundaries" paragraph that now names the skill to hand off to (`compose-multiplatform`, `swift-ios`, `react-native`) instead of saying "the Compose/native skill", and absorbed the sibling map from the old See Also.

### DIFF
--- a/skills/flutter/SKILL.md
+++ b/skills/flutter/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: flutter
-description: "Use when building, structuring, testing or optimizing a Flutter app (Dart 3, Riverpod 3 / Bloc, go_router, Material 3, freezed, widget & golden tests). Triggers: creating a Flutter feature, choosing state management, wiring navigation/auth guards, modeling async/error state, JSON/dio data layer, killing widget rebuilds/jank, or writing widget/golden/integration tests. Stack: Flutter 3.44 / Dart 3.12."
+description: "Use when building, structuring, testing or optimizing a Flutter app — feature-first layering, Riverpod 3 or Bloc, typed go_router, freezed models, a dio data layer, Material 3, jank hunting, widget/golden tests. Targets Flutter 3.44 / Dart 3.12. NOT React Native (that is `react-native`), NOT Compose Multiplatform (that is `compose-multiplatform`)."
 tags: [flutter, dart, mobile, app, ios, android]
 recommends: [design, deployment]
 origin: risco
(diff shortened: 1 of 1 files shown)


ericrisco:

PR #107, opened 2026-07-26
refactor(rust): drop the duplicated rule bank and trigger list, keep the ownership teaching

ericrisco:
Context-engineering pass on `rust` against `scripts/skill-rubric.md`. Format, not content.

| | Before | After |
| --- | --- | --- |
| description | 786 chars | 346 chars |
| body | 20092 B / 396 lines | 14621 B / 321 lines |
| eval-lint | PASS | **PASS** (6 should_trigger / 5 should_not / 1 capability) |

## Description

The old one ended in a keyword list: compiler error strings, `.rs`/`Cargo.toml`, crate names, and
Catalan and Spanish phrasings — in context on every turn the skill is installed, whether or not it
fires. It also said `NOT ... tauri` only, which is not the boundary anyone is actually at risk of
crossing. The new one keeps what the skill does and adds the discriminating pair: `go` (the same
write/review/test/ship service shape, GC instead of ownership) and `tauri` (desktop webview shell).

## What went

- **"When to use / When NOT to use"** (23 lines) — every delegation bullet existed verbatim in the
  trailing sibling index, and the Use-when half restated the description.
- **"Decision rules"** (20 lines) — eight of ten rules reappeared below in the section that owns
  them or in the anti-patterns table (`?` over `.unwrap()`, thiserror vs anyhow, no guard across
  `.await`, `spawn_blocking`, bind your SQL, `forbid(unsafe_code)`, clippy as a gate). The two with
  no other home moved next to the step they govern: borrow-in/own-out into Ownership,
  "parse, don't validate" into the axum section.
- The inline **"Anti-patterns:"** paragraph under Errors — its three items are table rows.
- **Three table rows** whose fix a Bad/Good block directly above already demonstrated in the same
  words (Mutex held across `.await`, `format!` into SQL, `String` params).
- **Six "Quick reference" rows** that restated body idioms. What is left is the gates and commands
  you cannot get from the prose.
- The **tail reference index** and the verbose **See Also**. All four `references/` files keep an
  inline link where you would want them (catalog-diet's reachability invariant holds), and the
  sibling links now sit where the delegation is actually decided — `secure-coding` and `postgresdb`
  in Security, `deployment` in Production, `go`/`tauri`/`harness` at the foot.
- **"Project grounding"** collapsed 18 lines to 6 — same path, same contents to record, same
  "recorded, not gated" rule.

## What stayed, deliberately

The ownership section and its four-error catalog, the shared-state decision table, the 3-layer error
flow with `IntoResponse`, both async pitfalls as worked Bad/Good pairs, project layout, testing,
sqlx/`cargo audit` security, and the release profile. This skill's job is to front-load ownership;
that material is the reason it exists.

The **anti-patterns table** stays — the rubric requires one. It was a hybrid: part concrete failure
mode, part rationalization echo of the rule bank above it. With the rule bank gone the echo rows are
the only ones that could go, so it is retitled from "Anti-patterns / rationalizations -> STOP" and
its column reframed by failure mode rather than by excuse. 12 rows to 10.

## Substance untouched

Rust 1.85+ / edition 2024, tokio 1.x, axum 0.8 `{id}` syntax, thiserror 2, sqlx, tracing, every
command, every release-profile flag. No recommendation changed, nothing added.

`manifest.json` deliberately not touched — the orchestrator regenerates it.

### DIFF
--- a/skills/rust/SKILL.md
+++ b/skills/rust/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: rust
-description: "Use when writing, reviewing, testing, or shipping Rust code or async services - ownership/borrowing (move vs borrow vs clone, Arc/Rc/RefCell, lifetimes), error modeling with Result/?/thiserror/anyhow, async on tokio (.await, JoinSet, select!, spawn_blocking, Send-across-await), axum 0.8 HTTP handlers/extractors/State, cargo test (unit/integration/doctests), and sqlx + cargo-audit security. Triggers: \"write an axum/tokio service\", \"is this idiomatic Rust\", \"add tests\", compiler error text like \"value moved here\", \"borrow of moved value\", \"future cannot be sent between threads safely\", \"does not live long enough\", \"cannot borrow as mutable\", \"el borrow checker no em deixa\", \"petición async en Rust\", .rs files, Cargo.toml, thiserror, sqlx. NOT desktop/GUI shells (that is tauri)."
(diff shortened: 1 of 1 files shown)


ericrisco:

PR #110, opened 2026-07-26
feat(catalog): add viral-score, a 0-100 rubric for deciding which clips to publish

ericrisco:
Adds `viral-score`: ten weighted criteria, seven stacking penalties, a clamped 0-100 result and a Publish/Review/Discard verdict, plus ranking and an explicit cut-off when scoring a batch.

**The rubric is reproduced exactly** — same weights, same penalty values, same bands, same required output fields. Its value is that it is fixed: two clips are comparable, and the same clip scores the same twice. The body instructs the agent not to substitute its own judgment or retune the weights.

## Two deliberate departures from the text as supplied

Both easy to revert if you disagree:

**1. The description was in the pre-Claude-5 style.** It carried a `Dispara con: "valora este reel", "puntúa estos clips"…` trigger list — exactly what the rubric change in #23 removes, and what ~85 skills have been migrated away from today. Adding it as written would have shipped a regression on the same day. Rewritten to what/when/NOT-boundary: **712 → 434 chars**.

**2. The channel-context section named a handle and two workspace-local paths.** This package is public on npm and the number cannot be recycled, so `@niundiaazero`, `03-REELS/_research/…` and `01-TOOLS/CLIPS/…` are out. The substance they carried is kept, generalised: *the excerpt must start on a sentence boundary, because where you cut decides the hook* now sits in "Applying it well", where it applies to anyone.

If this was meant as a personal skill rather than a catalog one, that changes both calls — say so and I will restore the original framing and keep it out of the public catalog.

Body is in English to match the other 257 skills. Rubric semantics unchanged.

## Verification

- `eval-lint` **PASS** — 7 should_trigger / 5 should_not_trigger / 2 capability
- `npm run validate` OK; description 434 chars, well inside the 1024 hard limit
- All five `route_to` siblings verified to exist (`shortform-packaging`, `shortform-ideation`, `shortform-editing`, `shortform-strategy`, `youtube-thumbnails`)
- Anti-patterns table present, as the rubric requires

The two capability cases target the failure this rubric actually suffers in practice: penalties silently skipped, and batch mode not producing a cut-off.

Note: `manifest.json` is not regenerated here — it is derived, and the ~80 open migration PRs would all collide on it. Regenerate once centrally after merging.

### DIFF
--- a/skills/viral-score/SKILL.md
+++ b/skills/viral-score/SKILL.md
@@ -1,0 +1,122 @@
+---
+name: viral-score
+description: "Use to judge whether a short-form clip is worth publishing — scores a reel, short or podcast excerpt 0-100 on ten weighted criteria with automatic penalties, then ranks a batch and says where the cut-off falls. Run it on transcript candidates before rendering. NOT writing the hook or caption (that is `shortform-packaging`), NOT inventing the idea (that is `shortform-ideation`), NOT executing the edit (that is `shortform-editing`)."
+tags: [viral, score, rubric, shortform, reels, shorts, tiktok, ranking, evaluation, retention]
+recommends: [shortform-packaging, shortform-ideation, shortform-editing, shortform-strategy, video-shorts]
+profiles: []
+origin: risco
+---
+
+# viral-score — is this clip worth publishing?
+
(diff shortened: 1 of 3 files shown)


ericrisco:

PR #115, opened 2026-07-29
refactor(document-processing): migrate to the Claude-5-era authoring rubric

ericrisco:
## What this is

Context-engineering pass on `skills/document-processing/SKILL.md` per `scripts/skill-migration-brief.md`. Format only — no recommendation, version, price, command, or figure changed.

| Metric | Before | After |
|---|---|---|
| `description` chars | 813 | **350** |
| Body bytes | 10,833 | **9,818** |
| Body lines | 214 | **206** |
| eval-lint | PASS | **PASS** (should_trigger=7, should_not_trigger=5, capability=1) |
| `scripts/verify.sh` | pass | **pass** (0 failures, 0 warnings) |

## What went

- **The `Triggers:` phrase list** (nine quoted prompts across English, Spanish and Catalan). The model matches by meaning; that list was paid for in context on every turn the skill is installed, invoked or not. It accounted for most of the 463 chars removed from the description.
- **The trailing `Handoffs:` bullet list.** It restated routing the intro boundary test already gives. Its one unique clause — this skill *produces* the text `rag` ingests, it does not index it — is folded into that intro line so the substance survives in exactly one place.
- **The standalone "never import `PyPDF2`" rule paragraph**, moved into the `pypdf` row of the engine-selection table, next to the step it governs.
- **The restated "why `auto_regenerate=False`" explanation.** The reason is already a comment on the line that sets it, and an anti-patterns row. The unique fact (checkbox/radio value is the field's `/V` on-state, often `/Yes`, not `True`) stays.

## What stayed, deliberately

- **The anti-patterns table** — concrete failure modes (OCR'ing a digital PDF, silent AGPL exposure, trusting OCR'd totals), not a rule bank restating the flow. The rubric requires one.
- **The engine-selection decision table** — the flow genuinely branches on goal, cost, privacy and licence.
- **Step 0 (text-layer check) and its code** — the most expensive mistake in this skill is OCR'ing a PDF that can be extracted losslessly.
- Every runnable recipe (pdfplumber, pypdf, docxtpl, ReportLab, Docling, Mistral OCR), the PyMuPDF AGPL warning, and the OCR digit-corruption warning with its `1.234,50` example.
- The inline link to `references/engines.md`, the package's only reference file — still linked, so still loadable.

Boundaries kept and verified against `skills/`: `structured-extraction`, `e-signature`, `spreadsheet-ops` all exist, as do the `recommends` entries `rag` and `data-scraper`. The nearest sibling is `structured-extraction`; the new lead sentence discriminates on the deliverable's shape (document bytes / literal content vs a schema-typed object), which is the same test the body opens with.

## Not touched

`manifest.json` (derived — the orchestrator regenerates it), `evals/`, `scripts/`, `references/`, and every other skill. `npm test` / `npm run validate` were not run: no `node_modules` in this worktree, per the brief.

### DIFF
--- a/skills/document-processing/SKILL.md
+++ b/skills/document-processing/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: document-processing
-description: "Use when you need to get content out of a document or build a document out of data — extracting text/tables from PDFs, filling interactive PDF forms (AcroForm), generating PDF/DOCX from templates, or OCR'ing scanned/image-only files into Markdown/JSON. Triggers: 'extract the line items from this invoice PDF', 'fill this PDF form and flatten it', 'generate a contract DOCX from this template', 'this scan has no text layer', 'the PDF copies out as garbage symbols', 'merge/split/rotate these PDFs', 'OCR 300 scanned pages', 'rellena este formulario PDF y fusiónalos', 'extreu les taules d'aquesta factura escanejada'. NOT pulling schema-typed fields out of text (that is structured-extraction), NOT sending it for signature (that is e-signature), NOT reading spreadsheet cells/formulas (that is spreadsheet-ops)."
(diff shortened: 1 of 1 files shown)


ericrisco:

PR #116, opened 2026-07-29
refactor(docker): migrate to the Claude-5-era authoring rubric

ericrisco:
Context-engineering pass on `skills/docker/SKILL.md` per `scripts/skill-migration-brief.md`.
No substance changed: image names, tags, flags, sizes and version claims (Node 24 LTS, Go
1.26/1.25, Compose 2.22.0 `watch` GA, Chainguard/Wolfi SLSA L3) are byte-identical.

## Numbers

| | before | after |
|---|---|---|
| description | 655 chars | **347 chars** |
| body | 13171 bytes / 283 lines | **11062 bytes / 251 lines** |

Diff is 12 insertions / 44 deletions in one file.

## What went

- **The `Triggers: '…'` list** (eight quoted phrases across EN/ES/CA). It was keyword
  matching for a model that matches by meaning, charged to every turn the skill is
  installed. The description keeps what it does, when to use it, and three `NOT … (that
  is <sibling>)` boundaries: `deployment`, `scaling`, `secure-coding` — all real skills.
- **`## When to use`** — five bullets restating the description a screen below it.
- **`## The five rules`** — each rule was already in the anti-patterns table and
  demonstrated in the skeletons. Classic repeat-across-layers. Nothing was lost: the
  mechanisms only that section carried were moved to the line they govern (`/bin/sh -c`
  swallowing `SIGTERM`; digest pin reproducibility; PID 1 = UID 0; HEALTHCHECK readiness
  gating, now its own row; one-concern-per-container, into the Compose lead-in).
- **`## References & siblings`** — a tail index whose reference links already appear inline
  at point of use. The unique detail it carried (Rust / JVM `jlink` / static SPA → nginx
  templates, multi-arch buildx) moved into the inline mention, so the reader still knows
  what loading that file buys.
- The hand-off paragraph's second sentence, which repeated row 1 of the ownership table
  in the same words.

## What stayed, on purpose

- **The anti-patterns table** — the rubric requires one, and it is a different animal from
  a rule bank: bad/good pairs with the mechanism. It is now the single home for the
  layer-order and root rules rather than their duplicate.
- **The ownership table** — five owners route out of this skill, and three of them
  (`coolify`, `railway`, `fly-io`) appear nowhere in the description, so cutting it would
  lose real routing.
- **The base-image comparison table, the three multi-stage skeletons with their
  before/after sizes, the `.dockerignore`, the hardening `docker run` flags, the BuildKit
  secret mount, the Compose `watch` example, and the verify list.** This is the skill.

## Invariants checked

- Both `references/` files still linked from the body (`base-images-and-stages.md` line 50,
  `compose-recipes.md` line 225).
- Every `../<sibling>/SKILL.md` link resolves to an existing skill.
- No `npm test` / `npm run validate` run — no `node_modules` in this worktree, per the brief.

### DIFF
--- a/skills/docker/SKILL.md
+++ b/skills/docker/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: docker
-description: "Use when authoring or auditing a Dockerfile/Containerfile, shrinking a bloated image, hardening a container that runs as root, choosing a base image, or wiring a Compose dev loop with hot reload. Triggers: 'dockerize this app', 'my node image is 1.2GB make it smaller', 'the build cache keeps busting', 'harden this Dockerfile, trivy flags it', 'docker compose watch for local dev', 'distroless vs alpine vs slim', 'la imagen pesa demasiado', 'el build de docker tarda siglos'. NOT the CI/CD pipeline or deploy-to-host (that is deployment), NOT k8s autoscaling/HPA (that is scaling), NOT app-level SQL injection or secrets-in-code (that is secure-coding)."
+description: "Use when authoring or auditing a Dockerfile, shrinking a bloated image, hardening a container that runs as root, picking a base image, or wiring a Compose dev loop with hot reload. NOT CI builds or deploy-to-host (that is `deployment`), NOT k8s autoscaling (that is `scaling`), NOT app-level injection or secrets-in-code (that is `secure-coding`)."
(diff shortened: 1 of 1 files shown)


ericrisco:

PR #118, opened 2026-07-29
refactor(digitalocean): cut the trigger list and the duplicated tails

ericrisco:
Context-engineering pass on `skills/digitalocean/SKILL.md` against `scripts/skill-rubric.md`. No content rewrite: every recommendation, command, price and date is unchanged.

| | before | after |
| --- | --- | --- |
| description | 828 chars | 349 chars |
| body | 12,908 bytes / 251 lines | 11,403 bytes / 227 lines |

## Description

Was a keyword bank: eight quoted trigger phrases in three languages plus a feature inventory, in context on every turn the skill is installed whether or not it fires. Now what it does / when / an explicit NOT boundary against the three siblings a reader genuinely confuses this with:

- `deployment` — host-agnostic CI/CD and rollback strategy
- `hetzner` — a bare VPS you harden yourself
- `railway` — someone else's managed PaaS

The non-obvious "S3-compatible bucket with a CDN → Spaces" route is preserved as `S3-compatible Spaces + CDN` in the description and stays covered by its `should_trigger` eval case.

## Removed

- **`## References` tail index** — both files were already linked at point of use (`references/app-spec.md` in the App Platform section, `references/droplet-ops.md` in the Droplets section), so the index was a second copy of the same two links. Its one unique detail (VPC + private-DB layout) folded into the inline mention, so both links survive.
- **`## Boundary` section** — restated the description's NOT clause plus the `postgresdb` / `monitoring` / `scaling` delegations that already sit inline in the sections that need them. The routing it uniquely carried (`docker`, `domains-dns`) moved to the opening paragraph, where a reader decides whether they are in the right skill at all. This also resolved the dangling "see the boundary at the bottom" pointer.
- **A duplicated `bash` fence** under "Basic ops + cost gotchas" — both commands in it (`doctl apps logs`, `doctl apps update --spec`) appear verbatim in the App Platform section above. The cost bullets, which carry the real numbers, stayed.
- **Two rules-of-thumb bullets** collapsed into the ones beside them; the "App Platform components are not meant to be a persistent pet process" why is already the second column of the matching anti-patterns row.

## Kept

- **The anti-patterns table** — the rubric requires one, and this is not a rationalization bank: eight concrete failure modes with real slugs and prices (`apps-d-2vcpu-4gb` ~$78/mo, plaintext `dop_v1_…` in a spec, powering off a Droplet to stop billing).
- **The Droplet vs App Platform vs Functions decision table** — this is where the flow actually branches, and it is the skill's whole job.
- **Every load-bearing figure and date**: per-second billing since 2026-01-01 (min 60s / $0.01), 3 free static sites, $5/mo basic, ~$15/mo single-node managed PG and ~$30/mo HA, ~21 pools / ~1,000 connections, $5/mo Spaces with 250 GiB + 1 TiB egress, $0.02/GiB App Platform overage.
- **The app spec YAML, the `${db.DATABASE_URL}` injection and `type: SECRET` scoping notes, the boto3 Spaces snippet, and the "rollback = re-apply the prior spec" mechanism.**

No result envelope or chain-position block existed, so none was added. `manifest.json` untouched; `evals/`, `scripts/verify.sh` and both `references/` files untouched.

### DIFF
--- a/skills/digitalocean/SKILL.md
+++ b/skills/digitalocean/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: digitalocean
-description: "Use when deploying or operating a workload on DigitalOcean and you must decide where it runs and how to ship it — Droplet (raw VPS) vs App Platform (managed PaaS) vs Functions, wiring a Managed Database (Postgres/MySQL/Valkey), Spaces object storage + CDN, doctl auth, App Platform app spec YAML, cloud firewalls, VPC private networking, reserved IPs, snapshots/backups, app logs, scaling. Triggers: 'deploy to DigitalOcean', 'DigitalOcean App Platform app spec', 'DO managed postgres with connection pooling', 'doctl apps create', 'reserved IP failover', 'I need an S3-compatible bucket with a CDN' (that is Spaces — non-obvious), 'subir mi web a DigitalOcean con doctl', 'Droplet o App Platform, qué sale más barato'. NOT host-agnostic CI/CD or rollback strategy (that is deployment), NOT a Hetzner bare VPS (that is hetzner)."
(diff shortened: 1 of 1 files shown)


ericrisco:

PR #121, opened 2026-07-29
refactor(deep-learning): drop the trigger list and the guardrail/checklist duplication

ericrisco:
Context pass on `deep-learning` against `scripts/skill-rubric.md`. Format only — no
recommendation, version, figure, API name or code line changed.

| | Before | After |
|---|---|---|
| `description` | 1012 chars | **360 chars** |
| Body | 16206 B / 263 lines | **13818 B / 226 lines** (-15%) |
| eval-lint | — | **PASS** (8 should_trigger / 5 should_not_trigger / 1 capability) |

### Removed

- **The `Triggers:` list** — eight quoted phrases in English and Catalan, plus a
  `Covers …` inventory that repeated the section headings. That text sat in context on
  every turn the skill was installed, invoked or not, and the model matches by meaning
  anyway. The new description keeps what discriminates: the loop it owns, and the three
  `NOT` boundaries against `finetuning` (pretrained LLM + LoRA), `machine-learning`
  (tabular) and `nlp` (tokenization/metrics).
- **`Guardrails (the silent killers)`** — its `zero_grad`, raw-logits and
  fp16-needs-a-scaler bullets were already stated next to the lines they govern in §2
  and §3. §2 said so out loud: *"all in the gotchas table, worth stating once here"*.
- **`Related skills`** — a four-item tail index of links the routing table at the top of
  the file already carries.
- **The 10-item checklist** — every item restated §2–§6 one bullet at a time; each was
  checked against the body before cutting, and nothing in it was unique to the checklist.

### Kept

- **An anti-patterns table**, rebuilt from the Guardrails items that had no in-flow home:
  validating without `eval()`/`no_grad()`, NaN-loss triage, fp16 with the scaler off, a
  CPU tensor in a CUDA graph, and over-promising bit-exact seeds. The rubric requires
  one, and symptom-first failure modes are a different animal from restated rules.
- **The `Am I in the right skill?` routing table** — a genuine branch. It picked up the
  `python` link that would otherwise have died with the tail index, so all four
  `recommends` siblings stay linked.
- **The bf16-vs-fp16 decision table**, every code block, the DDP/FSDP2/ZeRO when-to-use
  lines, the quoted `randomness.html` note, and the version hedge (PyTorch 2.x, ~2.13
  mid-2026, don't hard-pin a minor).
- Both references stay linked inline — `references/distributed.md` from §5,
  `references/training-recipes.md` from the reproducibility section.

`manifest.json` untouched; the orchestrator regenerates it after merge.


ericrisco:

PR #122, opened 2026-07-29
refactor(error-handling): cut the rule bank, the tail decision table and the restated checklist

ericrisco:
Context-engineering pass on `skills/error-handling/` against `scripts/skill-rubric.md`, per `scripts/skill-migration-brief.md`. Format only — no technical claim, number, version or command changed.

| | Before | After |
|---|---|---|
| `description` | 771 chars | **351 chars** (-54%) |
| `SKILL.md` | 14845 B / 307 lines | **12778 B / 273 lines** (-14% / -34 lines) |
| `eval-lint.sh` | — | **PASS** (7 should_trigger / 5 should_not_trigger / 1 capability) |

## Description

Deleted the 9-phrase `Triggers: '…'` list (EN/ES/CA) — the model matches by meaning, and that list was in context on every turn error-handling is installed, invoked or not.

Kept all three `NOT` boundaries, because they are the whole discrimination problem for this skill: **`debug`** (one crash vs a class of failures), **`observability`** (telemetry pipeline vs runtime reaction), **`api-design`** (wire envelope vs the in-process status this skill produces). All three exist under `skills/`. The lead was rewritten to "the reaction to a class of failures", which is the phrase that separates it from `debug` in one read.

## What went

- **"The one rule"** (12 lines) — a rule bank that announced the flow and then said "do them in order". Each of its three bullets restated the step below it: *Classify* is Step 1's "the bucket dictates the reaction", *Contain* is Step 3's "no timeout is a bug, not a default", *Surface* is Step 5's two-audience table. The rule survives as one sentence in the opening, where the H1 already names it.
- **"Decision table"** (13 lines, 8 rows) — a tail quick-reference. Every row was already stated at the step that decides it: `Result` vs `throw` in Step 1, the retry and idempotency-key rows in Step 2, circuit-break + fallback in Step 3, `error.tsx` / event-handler / process-handler in Step 4.
- **"Verification checklist"** (13 lines, 8 boxes) → **"The gate"** (6 lines). All eight boxes duplicated the anti-patterns table immediately above them, which is already a checkable list carrying the failure mode *and* the fix per row. The `scripts/verify.sh` pointer, its advisory posture and `--strict` survive intact.

## What stayed, deliberately

- **The anti-patterns table, all 14 rows.** The rubric requires one, and this is not a rationalization bank: each row names a concrete failure mode ("thundering herd", "double-charges", "logs are a dead end") with its fix. It is now also the exit checklist.
- **The decision tables inside Steps 1, 2 and 5** — bucket → reaction, transient vs permanent, bucket → HTTP status. That is where the flow genuinely branches.
- **Both bad/good pairs** (string `throw` vs typed class with `cause`; leaked `TypeError`/`ECONNREFUSED` vs `503` + correlation id) and the `neverthrow`, `withRetry` and `error.tsx` skeletons — they teach by demonstration.
- **Every non-obvious hard directive, next to the step it governs, each with its reason**: `error.tsx` must be a Client Component; `error.tsx` does **not** catch its own segment's `layout`/`template` error; boundaries catch render errors only; `global-error.tsx` renders its own `<html>`/`<body>`; per-attempt timeout *before* the retry; idempotency key before retrying a mutation; exit and let the supervisor restart. Nothing moved to an appendix — a rule at the step is followed, a rule in an appendix is skimmed.
- **Substance verbatim**: three buckets, retry/never-retry table, caps (3–5 attempts, 100–200ms base doubling, 10–30s per-delay cap, 10–60s budget, AWS Builders' Library / REL05-BP03), the full-jitter formula, breaker thresholds (~50% over ~20 reqs, Open 30–60s, ~30% critical / ~70% tolerant), Opossum's defaults, Polly 8.6.6, Resilience4j 2.3.0, the taxonomy → status map.
- **Both `references/` files, still linked inline at point of use** — `retry-and-resilience.md` from Steps 1–3, `boundaries-and-messaging.md` from Steps 4–5. An unlinked reference is never loaded.

## One non-format change

`chmod +x skills/error-handling/scripts/verify.sh`. It was mode `100644` while 186 of the catalog's 189 `verify.sh` files are `100755`, so the rubric's "verify.sh is executable" gate could not pass. Contents untouched; it still exits 0 on a clean target.

## Verified

- `bash scripts/eval-lint.sh` → `error-handling  PASS (should_trigger=7, should_not_trigger=5, capability=1)`
- description 351 chars, frontmatter parses as YAML; `name` matches the directory; `origin: risco` present
- all 7 `recommends` ids resolve to real skills; all 4 `../<sibling>/SKILL.md` links resolve
- both `references/` files linked from the body
- `manifest.json` untouched (derived — the orchestrator regenerates it once at the end)
- `npm test` / `npm run validate` not run: no `node_modules` in this worktree, per the brief


ericrisco:

PR #126, opened 2026-07-29
refactor(embeddings-search): trim description to a 350-char boundary, drop the duplicated tail index

ericrisco:
Context-engineering pass on `embeddings-search` against `scripts/skill-rubric.md`. Format only — no recommendation, model, dimension, price, figure or command was changed.

| | Before | After |
|---|---|---|
| `description` | 936 chars | **350 chars** |
| body | 11974 bytes / 194 lines | **10810 bytes / 184 lines** |

## What went

- **The `Triggers:` list** — eight quoted phrases across English and Catalan. The description is in context on every turn the skill is installed, invoked or not; the model matches by meaning, so a keyword list is cost with no routing gain.
- **The `Symptoms:` clause** — five restatements of the "when" the sentence before it already covered.
- **The trailing `## References & siblings` index.** Both reference files are now pointed to inline where they are needed — `references/models.md` in section 1 (MTEB over-trust caveat), `references/evaluation.md` in section 5 (metrics) — so every link survives and the tail was a second copy. The siblings line under it repeated the routing block that opens the body.
- **One intro sentence** ("when the symptom is…") that restated the description.

The `NOT` boundary went from naming one sibling to two: `vector-db` (store operation: index tuning, quantization) and `rag` (the retrieve-to-answer loop). Those are the two skills a reader is genuinely choosing between, so discrimination improved while the description got 63% shorter. Both exist under `skills/`, as do the other two the body routes to (`structured-extraction`, `prompt-engineering`).

## What stayed, and why

- **The anti-patterns table, all ten rows.** Dimension 3 requires one, and these are concrete failure modes (chunk size in characters, same `input_type` both sides, cosine model queried with L2), not rules restated from above. No STOP/excuse framing was present to strip.
- **The model matrix and the MTEB anchors** — dims, max input, price and query/document asymmetry per model are the substance of section 1.
- **The two hard rules** on distance metric and query/document asymmetry. They already sit next to the step they govern and justify themselves ("each is a **silent** failure, no error, just worse results") rather than shouting non-negotiable — which is the shape the rubric asks for.
- **The chunking upgrade ladder, the RRF snippet, the reranker model names and figures**, and the measure-retrieval section.
- **No result envelope was added.** This skill never had one.

`manifest.json` untouched.


ericrisco:

PR #127, opened 2026-07-29
refactor(duckdb): cut the trigger list and the duplicate reference index

ericrisco:
Context-engineering pass on `duckdb` per `scripts/skill-migration-brief.md` + `scripts/skill-rubric.md`. Format only — no recommendation, version, figure or command changed.

## Numbers

| | before | after |
|---|---|---|
| description | 746 chars | **350 chars** (-53%) |
| SKILL.md | 11771 bytes / 208 lines | **10768 bytes / 202 lines** |
| eval-lint | — | **PASS** (`should_trigger=6, should_not_trigger=5, capability=1`) |

## Description

Dropped the eight-phrase `Triggers:` list (English/Spanish/Catalan) — pure per-turn cost for every install, invoked or not. What remains is what discriminates: in-process analytical SQL over files, plus the two `NOT` boundaries.

> Use when analytical SQL must run in-process with no server: Parquet/CSV/JSON/Arrow queried in place, OLAP embedded in an app or notebook, a slow pandas groupby on multi-GB data, or S3/lakehouse data read without downloading. NOT a multi-user analytics server (that is clickhouse-analytics), NOT an app's transactional CRUD store (that is postgresdb).

Both named siblings verified present under `skills/`.

## What went

- **Trailing `## References` index.** Both files it listed are already linked inline at their point of use (`references/python-and-interop.md` in the Python interop section, `references/remote-and-lakehouse.md` in the remote/lakehouse section), so the tail block paid twice for the same two links. Invariant holds: every file under `references/` is still reachable from the body.
- **A stale hedge** — "(Some recommended siblings above may not be built in this collection yet; the routing decision still holds.)" Checked the catalog: `clickhouse-analytics`, `postgresdb`, `sqlite-turso`, `sql` and `vector-db` all exist, so the caveat was simply wrong.

## What was reframed, not cut

The anti-patterns table was the hybrid case the brief describes. **All nine rows kept**, re-columned `Rationalization | Reality → STOP` → `Anti-pattern | Do instead`, urgency wrapper dropped. The rows name concrete failure modes (pandas-then-query, DuckDB behind a 200-user dashboard, unpinned prod version, `SELECT *` over 200 columns) — that is the anti-patterns table dimension 3 requires, not a rule bank.

## What was deliberately kept

- **"Is DuckDB the tool?" routing table** — the flow genuinely branches six ways, and routing is this skill's primary job.
- Version guidance (v1.5.3 stable, v1.4.4 LTS), install block, friendly-SQL dialect block, remote/lakehouse and export snippets, the "Outgrowing DuckDB" handoff to `clickhouse-analytics`/MotherDuck, the `Verify` section describing `scripts/verify.sh`, and the 02-DOCS grounding pointer (a pointer to `harness`, not a restatement of it).
- No result-envelope block existed and none was added.

## Checks

- `bash scripts/eval-lint.sh` → `duckdb PASS`
- Frontmatter parses as YAML; `name` matches the dir id; `origin: risco`
- All five `../<sibling>/SKILL.md` links resolve; all five `route_to` values in `evals/cases.yaml` name real skills
- Only `skills/duckdb/` touched

### DIFF
--- a/skills/duckdb/SKILL.md
+++ b/skills/duckdb/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: duckdb
-description: "Use when you need fast analytical SQL over local files (Parquet/CSV/JSON/Arrow) with no server to run, when embedding OLAP in an app or notebook, when a pandas groupby/merge on multi-GB data is too slow, or when reading data straight from S3/HTTP/a lakehouse. Triggers: 'query a folder of parquet files in place', 'aggregate a 5GB CSV on my laptop', 'replace this slow pandas groupby', 'run SQL over a dataframe in Jupyter', 'read these parquet files from S3 without downloading', 'out-of-core aggregate', 'analizar un CSV enorme sin montar un servidor', 'consultar parquet en local sense base de dades'. NOT a multi-user production analytics server (that is clickhouse-analytics), NOT your app's transactional CRUD database (that is postgresdb)."
+description: "Use when analytical SQL must run in-process with no server: Parquet/CSV/JSON/Arrow queried in place, OLAP embedded in an app or notebook, a slow pandas groupby on multi-GB data, or S3/lakehouse data read without downloading. NOT a multi-user analytics server (that is clickhouse-analytics), NOT an app's transactional CRUD store (that is postgresdb)."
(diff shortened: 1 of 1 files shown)


ericrisco:

PR #128, opened 2026-07-29
refactor(expo): migrate to the Claude-5-era authoring rubric

ericrisco:
Context-engineering pass on `skills/expo/` per `scripts/skill-migration-brief.md`. Format only — no recommendation, version, command, or figure was changed.

## Numbers

| | before | after |
|---|---|---|
| description | 725 chars | **335 chars** |
| body | 12789 bytes / 253 lines | **11184 bytes / 238 lines** |

## Description

Dropped the nine-phrase `Triggers:` list (English, Catalan, Spanish). It was paid for in context on every turn the skill is installed, invoked or not, and the body already routes semantically. What remains is what-it-does + when + two `NOT … (that is <sibling>)` boundaries against skills that actually exist: `react-native` and `flutter`.

## What went

- **"When to use / when NOT to use"** — the "Use when" paragraph restated the description. The "Do NOT use when" half carried genuine routing to six real siblings, so it was folded into a single line under the existing verb rule instead of deleted.
- **The trailing "See also" index** — both `references/` files were already linked inline at point of use, and the sibling list duplicated the description plus the verb rule. `scripts/verify.sh` kept its mention, moved beside the `eas.json` block it gates.
- **The "→ STOP" framing** and the `Rationalization | Reality` columns on the anti-patterns table.

## What stayed, deliberately

- **All ten anti-pattern rows**, re-columned to `Anti-pattern | Do instead`. They name concrete failure modes (hand-editing `Info.plist`, hardcoded `runtimeVersion`, expecting an OTA to carry a native dep), not rules restated from above — the rubric requires the table; only the borrowed urgency wrapper was the problem.
- **The decision-rules table** — the flow genuinely branches: managed/bare/Expo Go/dev build, and `fingerprint` vs `appVersion`.
- **Every load-bearing fact, verbatim**: SDK 55 (RN 0.83.1, React 19.2.0, 2026-02-25) and the SDK 56 beta line, builder toolchain, EAS plan limits and dollar figures with their 2026-06-02 verification date, all commands, the `eas.json` / `app.config.ts` / config-plugin / EAS Workflow blocks, the bad/good `runtimeVersion` pair, and the New-Arch upgrade checklist.
- **The 02-DOCS grounding block**, compressed to one paragraph. It names this skill's own wiki page and fields and points at `harness` rather than re-teaching it.

## Invariants

Both `references/` files remain linked from the body. No result envelope existed and none was invented. `manifest.json` and `evals/cases.yaml` untouched — the evals were already 7/5/1 with real `route_to` siblings.

### DIFF
--- a/skills/expo/SKILL.md
+++ b/skills/expo/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: expo
-description: "Use when building, shipping, or hot-patching a React Native app with Expo — EAS Build/Submit/Update, eas.json build profiles, config plugins, prebuild/CNG, the managed workflow, confirming you are on the New Architecture (mandatory since SDK 55). Triggers: 'eas build', 'eas.json', 'config plugin', 'expo prebuild', 'my OTA update isn't showing up on devices', 'runtime version mismatch', 'how do I add a permission without editing Xcode/Gradle', 'newArchEnabled is gone from app.json', 'com publico una actualització OTA sense passar per la store', 'subir la app a la Play Store con EAS'. NOT generic React Native components/navigation/native-module authoring (that is react-native), NOT a Dart mobile app (that is flutter)."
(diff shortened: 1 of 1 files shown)


ericrisco:

PR #130, opened 2026-07-29
refactor(elixir): cut the trigger-phrase list and the duplicated reference index

ericrisco:
Migrates `skills/elixir/SKILL.md` to the new authoring rubric
(`scripts/skill-rubric.md`, applied per `scripts/skill-migration-brief.md`).

| Metric | Before | After |
|---|---|---|
| `description` chars | 700 | 303 |
| body bytes | 12352 | 11740 |
| body lines | 217 | 209 |
| `scripts/eval-lint.sh` | PASS | **PASS** (6 / 4 / 1) |

## Description

The description is in context on every turn the skill is installed, invoked or
not, so length is a cost and never a credit. The four-language `Triggers:`
phrase list was the bulk of the 700 chars and bought nothing the model does not
already do by meaning-matching.

What survives is the discriminating part — what it does, when, and the boundary:

> Use when writing or refactoring Elixir/OTP on the BEAM — GenServers, supervision trees and restart strategies, pattern matching, mix projects and releases — or when processes misbehave (restart loops, mailbox growth, call timeouts). NOT a Phoenix web app, LiveView, Ecto or channels (that is `phoenix`).

`phoenix` is a real sibling (`skills/phoenix/`) and is the only skill this one is
genuinely confusable with, so it is the boundary worth spending characters on.

## What went, and why

- **The trailing `See references/...` index.** Both reference files were already
  pointed to inline at the point of use, so the tail was a second copy of the
  same two links. The extra topics it named (Task, Agent, ETS, choosing a
  behaviour) were folded into the inline mention, so no signal was lost and both
  files stay linked.
- **Two meta lead-ins** that described the document's format rather than teaching
  Elixir: *"Each rule below earns its place; the why is one line."* and
  *"Rules, each with its why:"*.
- **One duplicated paragraph**: *"Drop to raw processes only for a throwaway
  fire-and-forget…"* restated the paragraph three lines above it; merged into it.

## What stayed, and why

- **The anti-patterns table.** The rubric requires one, and these rows are
  concrete BEAM failure modes with a real "why it bites" (atom-table exhaustion
  from `String.to_atom/1`, unbounded `cast` mailbox growth), not rules restated
  from the prose above.
- **Both decision tables** — process vs. plain function, and the three restart
  strategies. The flow genuinely branches at both.
- **Every bad/good Elixir pair.** They teach idiom by demonstration, which is
  what the rubric asks for instead of a rule bank.
- **The `## Verify` section**, because `skills/elixir/scripts/verify.sh` exists.

## Substance untouched

Elixir v1.19.5 / v1.20-rc and the OTP version floors, the built-in `JSON`
claim, restart strategies and restart values, `config.exs` vs `runtime.exs`,
`mix release`, and every command are byte-identical. This was a context pass,
not a content rewrite.

`manifest.json` deliberately not touched — it is derived and regenerated
centrally after merge.

### DIFF
--- a/skills/elixir/SKILL.md
+++ b/skills/elixir/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: elixir
-description: "Use when writing or refactoring Elixir/OTP code — GenServers, supervision trees, processes, pattern matching, mix projects and releases — or when BEAM behaviour misbehaves (restart loops, mailbox growth, timeouts). Triggers: 'write a GenServer', 'design a supervision tree', 'restart strategy one_for_one', 'let it crash instead of try/rescue', 'mix release for production', non-obvious 'my process mailbox keeps growing', 'GenServer call timeout / restart loop', Catalan 'arbre de supervisio en Elixir / processos que peten', Spanish 'arbol de supervision / procesos en Elixir / que es let it crash'. NOT building a Phoenix web app, LiveView, Ecto schemas, controllers or channels (that is phoenix)."
+description: "Use when writing or refactoring Elixir/OTP on the BEAM — GenServers, supervision trees and restart strategies, pattern matching, mix projects and releases — or when processes misbehave (restart loops, mailbox growth, call timeouts). NOT a Phoenix web app, LiveView, Ecto or channels (that is `phoenix`)."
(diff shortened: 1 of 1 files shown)


ericrisco:

PR #132, opened 2026-07-29
refactor(domains-dns): cut the trigger list and the tail index

ericrisco:
Context-engineering pass on `domains-dns` against `scripts/skill-rubric.md`. Format only — no recommendation, version, command, RFC citation or figure changed.

## Numbers

| | Before | After |
| --- | --- | --- |
| `description` chars | 754 | **338** (≤350 target, 1024 hard limit) |
| Body bytes | 11242 | **9841** |
| Body lines | 146 | **131** (≤400 ceiling) |
| eval-lint | PASS | **PASS** (7 should_trigger, 5 should_not_trigger, 1 capability) |

## Description

Dropped the ten-phrase, three-language `Triggers:` list. It is in context on every turn the skill is installed, invoked or not, and the model matches by meaning. The rewrite keeps what discriminates — the record types, apex vs www, TLS, cutover — and states the boundary against the two siblings it is genuinely confusable with:

> Use when pointing a domain at a host or fixing broken HTTPS — delegating nameservers, writing A/AAAA/CNAME/ALIAS/MX/TXT/CAA rows, apex vs www, standing up auto-renewing TLS, and cutting nameservers over without downtime. NOT inbox placement or warmup (that is `email-deliverability`), NOT what runs behind the name (that is `deployment`).

Both siblings exist under `skills/`.

## What went

- **The `Triggers:` phrase list** (~410 chars of the description).
- **The "Verification checklist before you call it done"** — its five items mapped one-to-one, in order, onto the five commands in the fenced block directly above it. Its two criteria that were not pure restatement (complete chain with `notAfter` ahead; www and apex on one canonical host) are exactly what `scripts/verify.sh` already audits, so the checklist is replaced by a one-line pointer to that script.
- **The trailing "References & siblings" index.** `record-cookbook.md` and `tls-and-acme.md` were already linked inline at point of use; `verify-and-debug.md` and `scripts/verify.sh` now are too, in the Verify section. The route-out paragraph repeated the description's boundary, and its `cloudflare` / `vercel` / `email-deliverability` links already appear inline in the body.

**Reference invariant holds:** all three files under `references/` are linked from the body.

## What stayed, on purpose

- **The anti-patterns table, all nine rows.** The rubric requires one, and this is not a rule bank or a rationalization table — no STOP framing, no excuse/reality columns. Every row names a concrete failure mode, why it bites, and the fix.
- **Resolution order is debug order** (name → delegation → records → TLS) — the core teaching, and the reason the skill routes at all.
- **Two decision tables where the flow genuinely branches**: where to host the zone, and HTTP-01 vs DNS-01.
- **The apex bad/good record pair** — judgment taught by demonstration.
- **The TLS rules block** — each rule already sits at point of use and names the failure it prevents (CAA before issuance, fullchain vs leaf, staging vs the 50-cert limit).
- **The nameserver cutover sequence** and the record cookbook summary table.
- No result-envelope block existed and none was added.

## Verification

- `bash scripts/eval-lint.sh | grep -E '^domains-dns +'` → PASS
- `npm test` / `npm run validate` not run — no `node_modules` in the worktree, per the migration brief.

### DIFF
--- a/skills/domains-dns/SKILL.md
+++ b/skills/domains-dns/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: domains-dns
-description: "Use when pointing a custom domain at a host, fixing broken HTTPS, choosing apex vs www, or publishing DNS rows — registering/delegating a name, writing A/AAAA/CNAME/ALIAS/MX/TXT/CAA/SRV records, standing up auto-renewing TLS, or cutting nameservers over without downtime. Triggers: 'point example.com at Vercel', 'set up DNS for the domain', 'NET::ERR_CERT_AUTHORITY_INVALID', 'CNAME flattening / ALIAS at the apex', 'dig +trace still shows the old IP', 'need a CAA record for Let's Encrypt', 'publish the SPF/DKIM/DMARC/MX rows', 'apuntar el dominio', 'el certificado caducó y no renueva', 'configurar els registres DNS'. NOT inbox placement / warmup / spam-rate tuning (that is email-deliverability), NOT what runs behind the name (that is deployment)."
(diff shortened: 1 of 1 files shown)


ericrisco:

PR #133, opened 2026-07-29
refactor(electron): cut the trigger list from the description and merge the duplicate rule tables

ericrisco:
Migrates `skills/electron/SKILL.md` to the Claude-5-era authoring rubric. Format pass only — no recommendation, version, command or figure changed.

| Metric | Before | After |
|---|---|---|
| `description` chars | 821 | 350 |
| body bytes | 13,248 | 11,720 |
| body lines | 221 | 209 |
| `scripts/eval-lint.sh` | — | **PASS** (should_trigger=5, should_not_trigger=4, capability=1) |

## Description

The description is in context on **every turn the skill is installed**, invoked or not, so its length is a cost paid by every user. Most of the old 821 chars was an eight-phrase `Triggers:` list in English, Spanish and Catalan — the model matches by meaning, so the list bought nothing.

The new 350-char version leads with `Use when`, names the surface (process model, typed contextBridge IPC, `nodeIntegration`/`contextIsolation`/`sandbox`/CSP lockdown, signing and auto-update), and closes with two boundaries against **real** siblings: `tauri` for the Rust-backed native-webview shell, and `react` for the web UI inside the window. Both exist under `skills/`. The test is discrimination, not coverage — a reader can now pick this over either sibling from the description alone.

## What went, and why

- **The opening blockquote.** It restated the description, then restated "the governing rule" printed twelve lines below it. Its one unique claim — that renderer-to-OS escalation is the dominant failure mode — moved *into* the governing rule, so the reason now sits next to the thing it justifies instead of being asserted twice.
- **"— non-negotiable" from the security-baseline heading.** The section already earns it concretely: one why per flag, plus "flipping any of them to the insecure value is what `verify.sh` fails on". The shouting added nothing the linter does not already enforce.
- **The separate "Migration smells" table.** Four of its five rows duplicated the anti-patterns table directly below it. Merged into one **"Anti-patterns and migration smells"** table. Every distinct row and every distinct reason survives — including `new BrowserView(...)` → `WebContentsView` and the sharper `@electron/remote` framing ("sync main-object access = renderer→main RCE"), which was the better of the two wordings.
- **The stale `(../react/SKILL.md if it existed)`.** `react` is in the catalog; the link resolves.

## What stayed, deliberately

- **Every security rule, next to the step it governs, with its reason intact.** `nodeIntegration: false` / `contextIsolation: true` / `sandbox: true` on the `BrowserWindow` snippet; CSP via `onHeadersReceived` rather than `<meta>`; the `will-navigate` + `setWindowOpenHandler` lockdown; the `shell.openExternal` protocol allow-list; "never expose `ipcRenderer` across the bridge"; validate-every-arg in `ipcMain.handle`. These are the failure mode the skill exists for — none was moved to an appendix or compressed into a rule bank.
- **The anti-patterns table**, which the rubric requires, and the local-only vs remote-content branch table, which is a genuine fork in the flow.
- **Freshness claims verbatim**: Electron 42 stable / Chromium M148 / Node 24, latest-3-majors support window, Forge vs electron-builder decision table, and the sign → notarize → publish → auto-update ordering.
- **Both inline `references/` links** (`security-and-ipc.md`, `packaging-and-updates.md`), so neither reference file goes dead.

No result-envelope block existed and none was added. `manifest.json` untouched — it is derived and regenerated centrally after merge.

### DIFF
--- a/skills/electron/SKILL.md
+++ b/skills/electron/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: electron
-description: "Use when building, hardening, or shipping a cross-platform desktop app with Electron — wiring the main/renderer/preload process model, exposing OS capabilities to a web UI over typed IPC, locking down an insecure shell, or packaging with code signing and auto-update. Triggers: 'build a desktop app for Mac/Windows/Linux from my web app', 'let my renderer read a file safely', 'audit my Electron security (nodeIntegration, sandbox, CSP)', 'my preload exposes ipcRenderer and the renderer can require(\"fs\")', 'Squirrel.Mac won\\'t auto-update my app', 'migrate off @electron/remote and BrowserView', 'empaquetar mi app de escritorio para Windows y Mac con firma de código y auto-actualización', 'empaquetar app d\\'escriptori amb auto-actualització'. NOT a smaller Rust-backed shell on the OS native webview (that is tauri)."
+description: "Use when building, hardening, or shipping a cross-platform Electron desktop app — main/renderer/preload process model, typed contextBridge IPC, locking down nodeIntegration/contextIsolation/sandbox/CSP, or packaging with signing and auto-update. NOT a Rust-backed shell on the native webview (that is tauri), nor the web UI inside it (that is react)."
 tags: [electron, desktop, ipc, security, packaging, auto-update, code-signing]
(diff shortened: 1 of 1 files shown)


ericrisco:

PR #134, opened 2026-07-29
refactor(dynamodb): cut the trigger list and the duplicated step order

ericrisco:
Migrates `skills/dynamodb` to the Claude-5-era authoring rubric. Context pass only — no
recommendation, version, figure or command changed.

| metric | before | after |
|---|---|---|
| `description` chars | 879 | 348 |
| body bytes | 13296 | 12221 |
| body lines | 217 | 204 |
| `scripts/eval-lint.sh` | PASS | **PASS** (6 should_trigger / 5 should_not_trigger / 1 capability) |

## Description

The old one carried a nine-phrase `Triggers:` list in English, Spanish and Catalan — 531 of its 879
characters — sitting in context on every turn the skill is installed, invoked or not. The model
matches by meaning, so the list bought nothing. The new description keeps what it does, when to reach
for it, and the two boundaries that discriminate it from its nearest siblings:

> Use when modeling or operating a DynamoDB table: deriving partition/sort keys from access patterns,
> single-table vs table-per-entity, adding a GSI/LSI, on-demand vs provisioned capacity, or diagnosing
> hot-partition throttling. NOT relational schema/SQL/EXPLAIN (that is `postgresdb`), NOT
> aggregation-pipeline document modeling (that is `mongodb`).

## What went

- **"The non-negotiable order"** — a four-item rule bank restating the opening paragraph and the
  Step 1-4 headings for a third time. Its load-bearing reasons were not dropped, they moved next to
  the step each governs: "you cannot query what your keys do not express" now opens Step 1, the
  post-launch migration cost (new table + backfill + dual-write) opens Step 2, and "capacity mode is
  billing, not modeling" opens Step 4.
- **The trailing "References & siblings" index** — both reference links already appeared inline at
  point of use, so the index was a second copy of the same two links. Their descriptions were folded
  into those inline mentions; nothing about what each reference contains was lost, and both files
  remain linked from the body.
- **The trailing "verify.sh" section** — folded into Step 1, beside the artifact the script checks.

## What stayed, deliberately

- **The anti-patterns table, untouched.** Nine concrete failure modes, each with a why and a fix,
  already in `Anti-pattern | Do instead` shape with no excuse/STOP framing — the rubric requires it.
- **Both decision tables** (single-table vs table-per-entity; capacity mode by traffic shape) — the
  flow genuinely branches at both.
- **The "what you produce / what you refuse" contract** and the route-elsewhere map, which names six
  engines the description has no room for (`mysql`, `redis`, `vector-db`, `aws-essentials`,
  `deployment`, `secure-coding`).
- **Every figure verbatim** — 400 KB item cap, 1 MB page, 3,000 RCU / 1,000 WCU per partition, 20 GSI
  / 5 LSI quotas, the 10 GB LSI item-collection cap, RRU/WRU pricing and the Nov-2024 cut.

## Checks

- `bash scripts/eval-lint.sh` → `dynamodb PASS`.
- `references/access-patterns.md` and `references/capacity-and-limits.md` both still linked from the
  body.
- `../postgresdb/SKILL.md`, `../deployment/SKILL.md`, `../secure-coding/SKILL.md` all resolve.
- `manifest.json` untouched — it is derived and regenerated centrally after merge.

### DIFF
--- a/skills/dynamodb/SKILL.md
+++ b/skills/dynamodb/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: dynamodb
-description: "Use when designing or operating an Amazon DynamoDB table - modeling entities and relationships without joins, choosing partition/sort keys, deciding single-table vs table-per-entity, adding a GSI/LSI, picking on-demand vs provisioned capacity, or diagnosing throttling. Triggers: 'design a DynamoDB table', 'model one-to-many / many-to-many without joins', 'single-table design worth it?', 'query the same data by two different keys' (inverted GSI), 'writes throttled even though I'm under my provisioned capacity' (hot partition), 'on-demand or provisioned?', 'why does my Query only return some rows?' (1 MB page), 'diseña la tabla DynamoDB para estos patrones de acceso', 'modela aquesta relació amb clau composta a DynamoDB'. NOT relational schema / SQL / EXPLAIN / B-tree indexing (that is postgresdb), and NOT document modeling with aggregation pipelines (that is mongodb)."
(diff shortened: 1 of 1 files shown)


ericrisco:

PR #136, opened 2026-07-29
refactor(firebase): migrate to the Claude-5-era authoring rubric

ericrisco:
Context-engineering pass on `skills/firebase/SKILL.md`. No content changes — every recommendation, version, limit, command and security-rule example is byte-identical or preserved verbatim in a new position.

## Numbers

| | Before | After |
|---|---|---|
| `description` | 718 chars | **333 chars** |
| Body | 251 lines / 14596 bytes | **237 lines / 13370 bytes** |

The description is the expensive half: it sits in context on every turn the skill is installed, invoked or not. It now reads *what it does · when · an explicit NOT boundary* and nothing else.

## What went

- **The `Triggers: '…'` phrase list** — five English phrasings plus a Catalan and a Spanish one. The model matches by meaning; a keyword list in three languages is pure per-turn cost. Those exact phrasings already exist as eval cases (including the Catalan Security Rules prompt and the emulator/requires-an-index one), which is where they are actually tested.
- **The `## When to use / When NOT to use` "When to use" bullets** — six bullets restating the description one layer down.
- **A duplicated opening fact** — "Shape data for reads" and "you shape data for the read path" were the same claim stated twice, ten lines apart. Merged into one bullet, no claim dropped.

## What stayed, deliberately

- **Every "When NOT to use" routing target**, re-columned into a compact table: `postgresdb`, `supabase`, `dynamodb`, `mongodb`, `gcp-essentials`, `react` / `nextjs`. These are boundary discrimination, not restatement — only `supabase` appears in the description. The supabase row keeps its "most-confused sibling, same BaaS shape / different data model and rules language" note.
- **The anti-patterns table verbatim.** Not a hybrid case: it was already `Anti-pattern | Why it's wrong | Do instead` with no STOP / excuse framing to strip.
- **All substance:** the bad/good Security Rules pair, the Storage rules pair, the modular-vs-namespaced SDK pair, the 1 MiB document limit, the ~1 sustained write/sec hotspot and sharded counters, rules-are-not-filters, `get`/`list` split, App Check, Blaze-plan requirement, Web SDK v12 / Node.js 22 / functions v7, the emulator-does-not-enforce-composite-indexes gotcha, and the client-API-key-is-not-a-secret note.
- **No result-envelope block was added** — this skill never had one.

## Invariants checked

- All three `references/` files (`data-modeling.md`, `security-rules.md`, `cloud-functions.md`) remain linked inline at point of use.
- `../postgresdb/SKILL.md` and `../nextjs/SKILL.md` resolve.
- `manifest.json` untouched (derived — orchestrator regenerates).
- Only `skills/firebase/SKILL.md` is in the diff.

### DIFF
--- a/skills/firebase/SKILL.md
+++ b/skills/firebase/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: firebase
-description: "Use when building on Firebase — modeling Firestore data, writing or auditing Security Rules, wiring Auth, shipping Cloud Functions, storing files in Cloud Storage, or choosing modular Web/Admin SDK imports — and when symptoms appear like 'my database is open to the internet', 'rules reject my query', a counter stuck at 1 write/sec, custom-claim RBAC, or a callable function. Triggers: 'firestore.rules audit', 'denormalize for read paths', 'collection vs subcollection', 'my query works in the emulator but fails in prod with requires-an-index', 'set a custom admin claim', 'regles de seguretat de Firestore', 'mi base de datos está abierta a internet'. NOT managed-Postgres BaaS with SQL and RLS (that is supabase)."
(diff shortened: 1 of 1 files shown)


ericrisco:

PR #137, opened 2026-07-29
refactor(e-signature): cut the trigger list, fold the rule bank into the steps it governs

ericrisco:
Context-engineering pass on `skills/e-signature/SKILL.md` per `scripts/skill-migration-brief.md` and `scripts/skill-rubric.md`. Format only — no recommendation, version, figure, command or legal claim was changed.

## Numbers

| | Before | After |
|---|---|---|
| description | 744 chars | **349 chars** |
| body | 12566 bytes / 185 lines | **12005 bytes / 185 lines** |

## What went

- **The `Triggers:` list** — ten quoted phrases in English, Spanish and Catalan. It is in context on every turn the skill is installed, invoked or not, and the model matches by meaning. The new description leads with "Use when", names both providers and the SES/AES/QES decision, and keeps an explicit `NOT` boundary against `contracts` (drafting the text) and `document-processing` (extracting fields) — both verified to exist under `skills/`.
- **The "Three rules" bank.** It restated the three sections directly below it. Each rule was moved next to the step it governs rather than deleted:
  - tier-first → the intro of *Decision: which legal tier do you need?*
  - sandbox / `test_mode` → *Auth & setup*, where the base path is actually configured
  - "a send is not done at `sent`" → the opening of *Webhooks / completion*
- **The boundary paragraph** that repeated the description in prose, compressed to one sentence that still carries both `../contracts/SKILL.md` and `../document-processing/SKILL.md` links.

## What stayed, on purpose

- **Every legal claim, word for word**: ESIGN Act + UETA has no tiers, eIDAS 2.0 / Reg (EU) 2024/1183 in force since May 2024, "a typed name (SES) is binding for most B2B in the US and EU", the AES/QES escalation and the hand-off to a lawyer. In this domain a trimmed legal statement is a changed one.
- The SES/AES/QES tier table and the DocuSign vs Dropbox Sign comparison — the flow genuinely branches on both.
- All SDK versions (`docusign-esign` 9.0.0, `@dropbox/sign` 1.11.0) and pricing figures, the env-var table, both send flows, the JWT Grant five steps, and the HMAC verification code.
- **The eight-row anti-patterns table.** The rubric requires one; this is a true `Anti-pattern | Why it is wrong | Do instead` table naming concrete failure modes, not a borrowed rationalization/STOP block, so it was kept as-is.
- The `Verify` section describing `scripts/verify.sh`, and the closing routes to `../gdpr-privacy/SKILL.md` and `../webhooks/SKILL.md`.

## Invariants checked

- All three `references/` files (`docusign.md`, `dropbox-sign.md`, `legal-tiers.md`) are still linked inline from the body.
- All four `../<sibling>/SKILL.md` links resolve.
- `name` matches the directory id, `origin: risco` intact, description parses as single-line quoted YAML at 349 chars.
- `manifest.json` untouched (derived — the orchestrator regenerates it).

### DIFF
--- a/skills/e-signature/SKILL.md
+++ b/skills/e-signature/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: e-signature
-description: "Use when adding legally-valid electronic-signature flows to documents — sending a PDF or template for signature, embedding signing in your own app, tracking status, verifying signing webhooks, and retrieving the signed PDF plus its audit trail via DocuSign or Dropbox Sign. Triggers: 'send this contract for signature', 'create a DocuSign envelope', 'signature_request/send', 'embed signing in our app', 'notify us when it's signed', 'set up DocuSign JWT auth', 'is a typed-name signature legally binding?', 'do we need QES for EU customers?', 'enviar a firmar', 'firma electrònica amb validesa legal'. NOT drafting the contract text itself (that is contracts) and NOT OCR/parsing/extracting fields from documents (that is document-processing)."
+description: "Use when wiring an e-signature flow with DocuSign or Dropbox Sign — picking the SES/AES/QES legal tier, sending a PDF or template for signature, embedded signing, verifying signing webhooks, retrieving the signed PDF plus audit trail. NOT drafting contract text (that is `contracts`), NOT extracting fields from PDFs (that is `document-processing`)."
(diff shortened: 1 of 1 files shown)


ericrisco:

PR #138, opened 2026-07-29
refactor(fly-io): Claude-5 rubric pass — description 565→343, drop the trigger list and the reference tail

ericrisco:
Claude-5-era authoring rubric pass on `fly-io`. Context-engineering only — no recommendation,
version, command, region code or figure changed.

## Numbers

| | before | after |
|---|---|---|
| description chars | 565 | **343** |
| SKILL.md bytes | 11747 | **11347** |
| SKILL.md lines | 206 | **197** |

## What went

- **The `Triggers:` phrase list** — seven quoted phrases in English and Catalan, ~220 chars in
  every user's context on every turn the skill is installed, invoked or not. The model matches by
  meaning; the list bought no routing it did not already have.
- **The `vercel` boundary** — the weakest of the old NOTs. Replaced by `deployment`, which is the
  sibling a reader is actually at risk of confusing this with (choosing a host vs operating on
  Fly). `railway` stays as the nearest-PaaS boundary. The `vercel` eval negative is untouched.
- **The six-line "Related skills" tail.** Three entries were already inline at point of use —
  `docker` at `[build]`, `postgresdb` in the anti-patterns row on unmanaged Fly Postgres,
  `secure-coding` after the secrets bad/good pair. The other three moved to where they route:
  `railway` in the opening paragraph, `domains-dns` beside the fly.toml field-surface pointer,
  `scaling` at the end of the scaling-lever section. Every link survives; all six resolve.

## What stayed, and why

- **The anti-patterns table.** Required by dimension 3, and its ten rows name concrete failure
  modes with a "why it bites" column — not a rationalization bank restating rules above it.
- **Both decision tables.** The region strategy branches on stateless vs volume-backed; the
  scaling table branches on autostop vs `scale count` vs `fly-autoscaler` vs `fly-replay`. Real
  forks, not decoration.
- **Both bad/good pairs** (shared-volume `[[mounts]]`, secret in `[env]`) — they teach the two
  gotchas that cost the most, by demonstration.
- **The Mental model section.** Five lines that make everything after them readable; the
  Volumes-are-pinned fact in particular is what the whole stateful half of the skill rests on.
- **The Verify section** for `scripts/verify.sh`.

## Checks

- `bash scripts/eval-lint.sh` → `fly-io PASS (6/4/1)`
- Both `references/` files still linked from the body: `fly-toml.md`, `multi-region.md`
- All six `../<sibling>/SKILL.md` links resolve
- Frontmatter parses; `name: fly-io` matches the dir, `origin: risco` present

### DIFF
--- a/skills/fly-io/SKILL.md
+++ b/skills/fly-io/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: fly-io
-description: "Use when deploying or operating an app on Fly.io — writing fly.toml, running Machines across regions near users, attaching Volumes, managing secrets, or scaling (autostop/autostart, scale count, fly-replay). Triggers: 'fly deploy', 'fly launch', 'desplega a Fly.io a prop dels usuaris', 'users in Sydney see high latency but my app only runs in iad', 'scale my Fly app to zero overnight', 'attach a persistent disk to my Fly machine', 'forward writes to the primary region'. NOT a generic git-push PaaS (that is railway), NOT frontend edge hosting (that is vercel)."
+description: "Use when deploying or operating an app on Fly.io — writing fly.toml, placing Machines in regions near users, attaching Volumes, managing secrets, or picking a scaling lever (autostop/autostart, scale count, fly-replay). NOT choosing which host to deploy on (that is `deployment`), NOT a git-push PaaS with no regions model (that is `railway`)."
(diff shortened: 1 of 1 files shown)


ericrisco:

PR #139, opened 2026-07-29
refactor(financial-model): migrate to the Claude-5-era authoring rubric

ericrisco:
Context-engineering pass on `skills/financial-model/SKILL.md` per `scripts/skill-migration-brief.md`. Format only — no recommendation, figure, formula, command or claim changed.

## Numbers

| | Before | After |
| --- | --- | --- |
| `description` | 1000 chars | 345 chars |
| Body | 13,227 bytes / 163 lines | 11,882 bytes / 145 lines |
| eval-lint | PASS | PASS (should_trigger=7, should_not_trigger=6, capability=1) |

## Description

Dropped the 12-phrase multilingual `Triggers:` list (EN/ES/CA). It is in context on every turn the skill is installed, and the model matches by meaning anyway. The replacement is what the skill does plus three explicit `NOT … (that is <sibling>)` boundaries against real siblings — `finance-ops` (the 13-week bank-reconciled view), `forecasting` (single-series statistical), `fundraising` (running the round) — so it discriminates rather than merely covers.

## What went

- **The duplicate routing.** The same destinations appeared three times: a "What it does NOT do — route out first" bullet list, three "STOP-and-route at the gate" bullets, and a "Hand-offs" tail. Collapsed into one **Route out first** table. All nine destinations (`finance-ops`, `bookkeeping`, `unit-economics`, `pricing`, `forecasting`, `pitch-deck`, `investor-materials`, `fundraising`, `cost-tracking`) and their reasons survive verbatim.
- **The trailing `references/` index.** All three reference files are already linked inline at point of use — `revenue-build.md` in Rule 2, `benchmarks-and-scenarios.md` in Rule 5, `model-structure.md` in Verify — so the invariant holds and the index was a second copy.
- One intro sentence that restated the routing boundaries.

## What stayed, on purpose

- **Rules 1–5 in full.** The burn/runway/cash-continuity formulas, the ~1.25–1.4 fully-loaded multiplier, the 75–80% GM bands, the 18–24 / 24–30 month raise-sizing numbers and the full 2025/26 benchmark table (burn multiple, Rule of 40, LTV:CAC, CAC payback, NRR, GM) with their sourced figures are substance.
- **The anti-patterns table**, all eight rows, re-columned to `Anti-pattern | Do instead` per the brief.
- The intake-gate table, the bottom-up/top-down decision table, the bad/good worked pairs, and the Verify section with the `verify.sh` contract.

No result-envelope block existed and none was added. `manifest.json`, `evals/`, `scripts/` and `references/` untouched — only `skills/financial-model/SKILL.md` is in this diff.


ericrisco:

PR #141, opened 2026-07-29
refactor(fundraising): migrate to the Claude-5-era authoring rubric

ericrisco:
Migrates `fundraising` to the Claude-5-era authoring rubric (`scripts/skill-rubric.md`, applied per `scripts/skill-migration-brief.md`).

## Numbers

| | Before | After | Δ |
| --- | --- | --- | --- |
| `description` chars | 950 | 353 | -63% |
| body bytes | 14328 | 12765 | -11% |
| body lines | 163 | 154 | -9 |

## Description

The old description ended in a 12-phrase `Triggers:` list in English, Spanish and Catalan. That list sits in context on **every turn the skill is installed**, invoked or not, and the model matches by meaning anyway — so it was pure per-turn cost. The new one is the rubric shape (what it does · when · explicit `NOT` boundary) and still discriminates against both nearest siblings by name:

> Use when planning or running an equity round as a process: sizing the raise to a milestone, choosing post-money SAFE vs priced round, tiering investors by intro path, sequencing outreach for momentum, or reading a term sheet before signing. NOT the slide narrative (that is `pitch-deck`), NOT the cap-table or valuation math (that is `financial-model`).

## What went

- **The `Handoffs + references` tail section.** All seven routes in it were already in the routing table at the top, and both `references/` files were already linked inline at the step that needs them — `funnel-math.md` in Step 3, `process-playbook.md` in Steps 2, 4 and 5. The invariant holds: every reference file is still linked from the body.
- **"You do NOT write the deck, build the model, or package the data room"** from the intro — the routing table immediately below says exactly that, with destinations attached.
- **The blockquote wrapper around the honesty rule** in Step 4. The rule moved inline into the paragraph it governs and now states *why* it is absolute rather than leaning on the quote block for volume.

## What stayed, and why

- **The anti-patterns table — all nine rows.** This is the hybrid case the brief calls out: it was titled "Anti-patterns / rationalizations → STOP" with an *Excuse → Reality* framing. Re-columned to `Anti-pattern | Do instead` and retitled; every row preserved. The rows name concrete failure modes (post-money SAFE pile-up, flat 200-VC blast, signing the first term sheet with no comp) — only the borrowed urgency wrapper went.
- **The intake gate.** A genuine branch: if the real ask is the deck or the model, it stops and routes instead of planning a round on unknowns.
- **The routing table, the SAFE-vs-priced decision table, the term-sheet bands table, and both Bad/Good pairs** — the flow really branches at each of those, and the pairs teach judgment by demonstration.
- **Every figure and citation verbatim.** Seed ~$3.1M at ~$16M pre-money · ~90% Carta Q1 2025 SAFE share, ~92% as of Q3 2025 · $15–25k priced-round legal cost · warm ~30–50% vs cold ~1–3% · ~98% 1x / ~95% non-participating · ~12.6% median seed lead ownership · 6–8 week sprint. Context-engineering pass, not a content rewrite.

No result-envelope block existed and none was invented.

## Verification

- `bash scripts/eval-lint.sh` → `fundraising PASS (should_trigger=8, should_not_trigger=7, capability=1)`
- All eight `../<sibling>/SKILL.md` links resolve; all seven `recommends` ids are real skills.
- `description` 353 chars, well inside the 1024 hard limit, parses as single-line quoted YAML.
- `manifest.json` untouched — it is derived and the orchestrator regenerates it after merge.

### DIFF
--- a/skills/fundraising/SKILL.md
+++ b/skills/fundraising/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: fundraising
-description: "Use when planning or running an equity fundraising round and you need the STRATEGY and PROCESS, not a single document — sizing the raise to a milestone, choosing post-money SAFE vs priced round, building and tiering the investor target list around intro paths, sequencing outreach to manufacture momentum, managing the diligence pipeline, or reading a term sheet to know what to push on before you sign. Triggers: 'plan our seed raise', 'how much should we raise and on what terms', 'SAFE or priced round, what cap', 'build my investor target list', 'how many investors do I actually need to contact', 'we have meetings but no term sheet — the raise stalled', 'got a term sheet, what do I push back on', 'planifica la ronda seed', 'cuántos inversores tengo que contactar', '¿SAFE o ronda con precio?', 'planifiquem la ronda d'inversió'. NOT the slide-deck narrative (that is pitch-deck) and NOT the cap-table/valuation math (that is financial-model)."
+description: "Use when planning or running an equity round as a process: sizing the raise to a milestone, choosing post-money SAFE vs priced round, tiering investors by intro path, sequencing outreach for momentum, or reading a term sheet before signing. NOT the slide narrative (that is `pitch-deck`), NOT the cap-table or valuation math (that is `financial-model`)."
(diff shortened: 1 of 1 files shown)


ericrisco:

PR #142, opened 2026-07-29
refactor(finetuning): cut the description 63% and fold the tail index inline

ericrisco:
Migrates `finetuning` to the Claude-5-era authoring rubric. Context pass only — no recommendation, version, hyperparameter or citation was changed.

## Numbers

| | Before | After |
|---|---|---|
| `description` | 981 chars | **364 chars** (−63%) |
| body | 16,615 bytes / 245 lines | **15,188 bytes / 233 lines** |
| eval-lint | — | **PASS** (should_trigger=7, should_not_trigger=6, capability=1) |

The description was the real cost: it is in context on every turn this skill is installed, invoked or not, and it was sitting at 981 of the 1024-char budget.

## What went

- **The `Triggers: '…'` phrase list** — 8 quoted phrases across English and Catalan. The model matches by meaning; a keyword list is pure per-turn cost.
- **The symptom enumeration** in the description ("when eval loss climbs while train loss keeps falling", "when the adapter seems to have learned nothing", "when inference outputs garbage after a run"). Each is already an anti-patterns row where it can be acted on. In the description it was coverage, not discrimination.
- **The `NOT … (that is vllm)` boundary.** Training vs serving is not a near-miss. The two boundaries that do real work are kept: `rag` (facts vs form — the fork that routes half of all "should I fine-tune?" questions) and `unsloth` (same job, one specific backend). `vllm` is still routed inline in the body and still has its own `should_not_trigger` eval case.
- **The `## Related skills` tail index** (12 lines). Every link in it already appeared at the point of use in the decision gate's route-out paragraph — except `huggingface` and the `data-cleaning` boundary. Those two, plus the "same LoRA/QLoRA concepts, one optimized implementation; this skill stays backend-agnostic" note that distinguishes this skill from `unsloth`, were folded into that paragraph. Nothing was lost, and a reader deciding where to go now reads it where they are deciding rather than 190 lines later.
- **The trailing `See references/…` line.** `methods.md` was already linked inline right after the GRPO example; `hyperparameters-and-eval.md` is now linked from the evaluation section it actually documents.

## What stayed, deliberately

- **Every substantive detail, verbatim.** QLoRA NF4 + double quantization, `alpha ≈ 2·r`, LoRA LR 1e-4–2e-4, DPO 5e-7, GRPO 1e-6, `target_modules="all-linear"`, `num_generations=8`, the trl v1.x / transformers v5.x version-reality notes and the `trl.experimental.*` import warning, LIMA's ~1,000 examples, all arXiv citations, and the model-license caveat. Nothing was reworded, re-recommended or re-versioned.
- **The anti-patterns table, unchanged.** It already had the correct shape (`Anti-pattern | Why it breaks | Do instead`) and names concrete failure modes rather than restating rules above it — there was no STOP / excuse framing to strip.
- **The decision gate and the method-selection tables.** The flow genuinely branches (fine-tune vs prompt vs RAG vs longer context; LoRA vs QLoRA vs full FT; DPO vs ORPO vs KTO vs GRPO by data shape).
- **The pre-flight checklist.** It guards an expensive, irreversible GPU run — operational, not a recap of prose.

## Invariants checked

- Both files under `references/` are linked from the body (`methods.md` line 160, `hyperparameters-and-eval.md` line 206).
- All three `../<sibling>/SKILL.md` links resolve (`rag`, `unsloth`, `vllm`).
- Every `recommends` id is still named in the body: `training-data`, `unsloth`, `open-weights`, `huggingface`, `vllm`, `rag`.
- `name` matches the directory id; `origin: risco` intact; description parses as single-line quoted YAML.
- Evals untouched and already healthy — all six `route_to` values name real skills in the catalog.
- `manifest.json` not touched.

### DIFF
--- a/skills/finetuning/SKILL.md
+++ b/skills/finetuning/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: finetuning
-description: "Use when adapting an open-weight model to a target FORM or BEHAVIOR — tone, output format, domain style, a reasoning/tool-calling pattern — via LoRA/QLoRA or full fine-tuning with TRL SFTTrainer and then preference optimization (DPO/ORPO/KTO/GRPO); when choosing fine-tune vs prompt vs RAG vs longer context; when setting LoRA r/alpha/target_modules; when eval loss climbs while train loss keeps falling; when the adapter seems to have learned nothing; when inference outputs garbage after a run. Triggers: 'fine-tune Llama/Qwen on our data', 'LoRA vs QLoRA vs full fine-tune', 'SFT then DPO/ORPO/KTO/GRPO', 'set target_modules / rank / alpha', 'my fine-tune overfits', 'the model forgot everything after training', 'write a GRPO reward function', 'ajustar un modelo amb LoRA amb les nostres dades'. NOT adding facts or fresh knowledge to a model (that is rag); NOT the fast single-GPU Unsloth backend or GGUF export (that is unsloth); NOT serving the trained model (that is vllm)."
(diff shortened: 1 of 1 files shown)


ericrisco:

PR #143, opened 2026-07-29
refactor(forecasting): drop the trigger list, keep every statistical claim

ericrisco:
Context-engineering pass on `forecasting` against `scripts/skill-rubric.md`.

| | Before | After |
|---|---|---|
| `description` | 635 chars | **346 chars** (-46%) |
| body | 11332 B / 139 lines | **11081 B** / 139 lines |
| eval-lint | PASS | **PASS** (7 should_trigger / 5 should_not_trigger / 1 capability) |

## What went

**The `Triggers:` list.** Seven quoted phrasings across English, Spanish and Catalan,
in context on every turn the skill is installed whether or not it fires. The model
matches by meaning, and every one of those phrasings already lives in
`evals/cases.yaml` as a `should_trigger` case — including the Spanish and Catalan
prompts and the lumpy/zero-sales demand case. Nothing that discriminated was lost.

**Restated prose in the accuracy bullets.** Three of the five bullets re-explained a
decision the deliverable contract or the loop had already stated, at length. They now
state it once. Every metric definition and threshold is intact: WAPE = total absolute
error / total actual volume, MAPE's blow-up as actuals approach zero, bias as
direction versus WAPE as magnitude, MASE scale-free with 1.0 as the pass/fail line,
rolling-origin over a single holdout, an interval always emitted.

## What stayed, on purpose

- **The deliverable contract** — script + `ds, forecast, lo, hi` artifact + accuracy
  readout. It already existed and `scripts/verify.sh` checks exactly those columns;
  the brief says do not remove an output contract that is already there.
- **The method-selection table.** AutoCES, AutoETS, AutoARIMA, CrostonOptimized and
  SeasonalNaive mapped to data shapes, with statsforecast v2.0.3 calls. Method names,
  the Croston 1972 intermittent-demand claim and the "< 2 seasonal cycles →
  SeasonalNaive only" rule are the substance of this skill, not padding.
- **Both code blocks**, including the zero-dependency pure-pandas seasonal-naive
  fallback — it is also the baseline every model must beat.
- The six-step loop, the six edge cases, and the handoff table.

## Also

- Anti-patterns table re-columned from `Bad | Good | Why` to the catalog's
  `Anti-pattern | Why it bites | Do instead`, matching `inventory`, `analyze`,
  `analytics` and `business-intelligence`. All seven rows unchanged; cells reordered.
- Both `references/` files are still pointed to from the body, now as real markdown
  links — `accuracy-and-backtesting.md` once, `methods-cheatsheet.md` twice.
- The `data-cleaning` handoff became a resolvable `../data-cleaning/SKILL.md` like its
  four siblings in that list.

## Gates

- Description 346 chars, single-line quoted YAML, `Use when` lead, two `NOT … (that
  is `sibling`)` boundaries naming `financial-model` and `inventory` — both real.
- All five `../<sibling>/SKILL.md` links resolve.
- Every `route_to` in `evals/cases.yaml` names a skill that exists.
- `manifest.json` untouched; `npm test` / `npm run validate` deliberately not run
  (no `node_modules` in the worktree).

### DIFF
--- a/skills/forecasting/SKILL.md
+++ b/skills/forecasting/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: forecasting
-description: "Use when you need to project sales, demand, units, revenue, signups or traffic forward from historical data and report a defensible number with an error band. Triggers: 'forecast next quarter's sales', 'how many units of SKU X next month', 'project revenue 12 months with a confidence band', 'is our forecast actually better than just repeating last quarter', lumpy/intermittent demand with lots of zero-sales weeks, 'prevé las ventas del próximo trimestre', 'prediu la demanda del proper mes per producte'. NOT building an assumption-driven P&L (that is financial-model), NOT sizing reorder points or safety stock (that is inventory)."
+description: "Use when projecting history forward — sales, demand, units, revenue, signups, traffic — into a defensible number with an error band: method by data shape, rolling-origin backtest, MASE vs the naive baseline. NOT an assumption-driven P&L or runway model (that is `financial-model`), NOT sizing reorder points or safety stock (that is `inventory`)."
 tags: [forecasting, demand-planning, time-series, sales-forecast, statsforecast]
 recommends: [inventory, financial-model, unit-economics, data-cleaning, analyze, duckdb]
 origin: risco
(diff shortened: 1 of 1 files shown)


ericrisco:

PR #149, opened 2026-07-29
refactor(hiring): migrate to the Claude-5-era authoring rubric

ericrisco:
Context-engineering pass on `hiring` against `scripts/skill-rubric.md`. Format only — no employment-law, validity, or figure changed.

## Numbers

| | before | after |
|---|---|---|
| description | 687 chars | 348 chars |
| SKILL.md | 12588 bytes / 248 lines | 11792 bytes / 236 lines |

## Description

The `Triggers: '…'` list carried six quoted phrases in two languages plus a `symptoms:` enumeration. That is paid for on every turn the skill is installed, invoked or not, and the model matches by meaning regardless. The replacement is built for discrimination rather than coverage: the four outputs this skill produces, the non-obvious "is our AI résumé filter legal" case, and two boundaries — `people-ops` for anything after the offer is accepted, `contracts` for offer terms. Both siblings exist under `skills/`, and both were already the body's stated hard boundary, so the description now agrees with it.

## What went

- The trailing `## References` index — `references/templates.md` was already linked inline at three points of use.
- `"Hiring gets the right person in the door; people-ops takes it from day one"` — a rhetorical restatement of the sentence directly above it.
- The `See "Write the job post"` pointer to the immediately following heading.

## What moved

- The **screen decision table**, out of a subsection of the funnel spine and into `## Screen the pile`, where the advance / work-sample / reject / parking-lot call actually happens. That retires the `"use the screen decision table above"` bridge sentence too.
- The screening-rubric pointer from the deleted tail index, now inline in the screen step — every file under `references/` stays linked from the body.

## What stayed, deliberately

- **The full 12-row anti-patterns table.** It is already `Anti-pattern | Why it fails | Do instead` with concrete failure modes, not a rationalization bank restating rules from above. The rubric requires one.
- **Every load-bearing claim verbatim:** NYC LL144, the EU AI Act Annex III high-risk classification and its 2 Dec 2027 date, Colorado's 2027 move, EEOC 1-year / 2-year retention, ban-the-box coverage, the Sackett et al. 2022 validity figures, and the 29% / 100%-vs-60% application-rate figures.
- Both worked bad/good pairs (job post, scorecard entry) and the scorecard skeleton — they teach the shape by demonstration.

`manifest.json` untouched; no `npm test` / `npm run validate` run (no `node_modules` in the worktree).

### DIFF
--- a/skills/hiring/SKILL.md
+++ b/skills/hiring/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: hiring
-description: "Use when there's an open role and you need to write the job post, screen inbound candidates, structure the interview loop, or build a scorecard to reach a fair hire decision; symptoms: gut-feel ratings, every interviewer asking the same question, a bloated requirements list, no consistent screen. Triggers: 'write a job description', 'screen these CVs', 'build an interview scorecard', 'structure the interview panel', 'redacta la oferta de empleo', 'necesito una scorecard', plus the non-obvious 'four people interviewed and we still can't decide' and 'is our AI résumé filter legal'. NOT managing the person after they're hired — onboarding, payroll, performance (that is people-ops)."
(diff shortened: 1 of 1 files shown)


ericrisco:

PR #150, opened 2026-07-29
refactor(observability): drop the trigger phrase list, keep the instrumentation body

ericrisco:
Claude-5-era rubric pass on `observability`. The body already met the rubric, so this is an honest small diff: the cost was almost entirely in the description.

## Numbers

| | before | after |
|---|---|---|
| description | 828 chars | **323 chars** |
| SKILL.md | 15,093 bytes / 232 lines | **14,346 bytes / 229 lines** |
| eval-lint | PASS | **PASS** (7 / 5 / 1) |

## What went

- **The `Triggers:` list — 505 chars, 8 phrases across English, Catalan and Spanish.** It is in context on every turn the skill is installed, invoked or not, and it restated what the first sentence of the description already said. The model matches by meaning; the keyword list bought nothing and was paid for every turn.
- **The second intro paragraph.** It restated the description's NOT clause with a "clean test" for the same monitoring-vs-observability split. The pointer to `../monitoring/SKILL.md` was folded into the first paragraph, so the routing survives without the duplicate.

The description now has the rubric shape — what it does, plus an explicit `NOT … (that is \`monitoring\`)` boundary. `monitoring` is a real skill under `skills/`.

## What stayed, and why

- **The entire technical body.** 229 lines against a 400-line ceiling, and every section earns its place: the three-pillars cost/alertability table, the OTLP + Collector architecture, the auto-instrumentation-before-manual ordering, the bad/good pairs for spans, logs and metric cardinality, the Collector pipeline, the alert list.
- **The anti-patterns table, untouched.** It was already `Anti-pattern | Why it bites | Do instead` — no STOP/excuse wrapper to strip, and no second near-duplicate rule table to merge. This skill only ever had the one.
- **Both `references/` files**, each still linked inline at its point of use: `instrumentation-recipes.md` (line 100) and `collector-config.md` (line 205). No unlinked reference.
- **No result envelope** — this skill never had one, so none was invented.

No recommendation, version, figure, endpoint or command was changed. This is a context-engineering pass, not a content rewrite.

## One correctness fix

`scripts/verify.sh` was missing its exec bit — one of only 2 of 189 `verify.sh` files in the catalog without it, and a deterministic gate 7 failure. Set to `100755`. Re-checked: it still exits 0 on an empty target.

## Verification

- `bash scripts/eval-lint.sh` → `observability PASS (should_trigger=7, should_not_trigger=5, capability=1)`
- Frontmatter parses as YAML; description 323 chars, well under the 1024 hard limit and under the 350 target.
- All `route_to` values in `evals/cases.yaml` (`monitoring`, `error-handling`, `performance`, `analytics`, `cost-tracking`) name real skills — checked, no stale "not in this catalog" claims here.
- `manifest.json` deliberately untouched.


ericrisco:

PR #151, opened 2026-07-29
refactor(machine-learning): Claude-5-era rubric pass — description 1017→460, body 260→228 lines

ericrisco:
Claude-5-era rubric pass on `machine-learning`. Context-engineering only — no recommendation, version, figure or command changed.

## Numbers

| | Before | After |
|---|---|---|
| `description` chars | 1017 | **460** |
| SKILL.md bytes | 19441 | **16989** |
| SKILL.md lines | 260 | **228** |
| `eval-lint` | — | **PASS** (should_trigger=7, should_not_trigger=6, capability=1) |

## Description

Dropped the seven-item `Triggers:` keyword list. It sat 1 KB from the top of the body, which already instructs semantic routing, and it is in context on every turn the skill is installed whether or not it fires.

Kept all four `NOT … (that is <sibling>)` boundaries — `deep-learning`, `data-cleaning`, `forecasting`, `nlp` — each verified to exist under `skills/`. They are what makes this skill pickable over its nearest neighbours, so they earn their chars. 460 is over the 350 aim; the four boundaries are the reason, and I judged discrimination worth more than the last 110 chars.

## What went

- **The `## Checklist` section, merged into the anti-patterns table.** The two overlapped almost completely — test set held out once, learned transforms inside the `Pipeline`, splitter matches the data, `Dummy` baseline, metric not accuracy on imbalance, target-leakage audit, train/test gap. That is the "repeating yourself across layers" pattern the migration targets. **No item was dropped to make it fit:** the two checklist items with no matching anti-pattern row (start from a clean validated table; set `random_state` and pin shipped versions) became new rows, so the table went 9 → 11.
- **`## Related skills`** — a second copy of the routing table at the top of the body.
- **The trailing `## References` index** — all three files are linked at their point of use, so the index was a third mention of links already in the body.

## What stayed, and why

- **The anti-patterns table itself**, re-columned `Rationalization | Reality → STOP` → `Anti-pattern | Do instead`. Every original row kept in substance; only the borrowed-urgency wrapper went.
- **The "is this the right skill?" routing table** — it genuinely branches seven ways, which is the case the rubric keeps decision tables for.
- **The dated version-reality block** and the Grinsztajn 2022 / arXiv:2207.08815 citation — dimension 4 is the heaviest and these are what carries it.
- **The metric-selection table** and the **worked end-to-end lifecycle** (baseline → CV → one test-set score), which teach by demonstration.
- **The 02-DOCS grounding note.**

## Invariants checked

- All three `references/` files still linked inline: `pipelines-and-cv.md` (Pipeline + CV sections), `gbdt-and-tuning.md` (GBDT section), `metrics-and-imbalance.md` (metrics section). The two pointers that only lived in the deleted tail — `python` for env/pinning, and importance/SHAP — were folded inline rather than lost.
- Every `../<sibling>/SKILL.md` link resolves.
- Frontmatter parses; `manifest.json` untouched.

## Incidental fix

`analytics`, `business-intelligence` and `forecasting` were named in bare backticks in the routing table. All three exist in the catalog, so they are now links like every other row.

### DIFF
--- a/skills/machine-learning/SKILL.md
+++ b/skills/machine-learning/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: machine-learning
-description: "Use when the task is a classic/tabular machine-learning problem — predict a column from rows of features with trees and scikit-learn: fit a RandomForest / XGBoost / LightGBM classifier or regressor, build a leak-proof Pipeline + ColumnTransformer, choose a cross-validation scheme (stratified / k-fold / time-series / grouped), pick the right metric on imbalanced classes (precision/recall/F1/ROC-AUC/PR-AUC, not accuracy), or debug a model that aced CV then collapsed in production. Triggers: 'train a model to predict churn/fraud/price', 'my ROC-AUC is 0.99 — is that leakage', 'which metric for imbalanced classes', 'cross-validation without leaking', 'XGBoost vs LightGBM', 'add a DummyClassifier baseline', 'entrena un modelo de clasificación tabular con scikit-learn'. NOT neural-net / PyTorch training (that is deep-learning), NOT cleaning the raw dirty table first (that is data-cleaning), NOT BI dashboards or forecasting a dated series (that is analytics/forecasting), NOT text/token modeling (that is nlp)."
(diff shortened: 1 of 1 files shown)


ericrisco:

PR #152, opened 2026-07-29
refactor(instagram-api): drop the trigger list, merge the two overlapping rule tables

ericrisco:
Context-engineering pass on `instagram-api` against `scripts/skill-rubric.md`. No substance changed — same recommendations, same v25.0 facts, same commands, same figures.

## Numbers

| | Before | After |
|---|---|---|
| `description` | 723 chars | **361 chars** |
| Body | 13425 B / 248 lines | **12567 B / 238 lines** |
| Anti-pattern rows | 7 (+5 in a second list) | **11 (one table)** |
| `eval-lint` | PASS | **PASS** (6 should_trigger / 4 should_not / 1 capability) |

## What went, and why

- **The `Triggers:` list.** Seven quoted phrases across English, Spanish and Catalan, ~330 chars of the description, in context on every turn the skill is installed whether or not it fires. The model matches by meaning; the keyword list is pure per-turn cost.
- **The five `**here**` rows in the routing table.** They restated the description word for word ("publish a Reel from a public `video_url`", "check the 24h publish cap"), and each already has its own body section.
- **The `## Migration & gotchas` bullet list.** A second rule table overlapping `## Anti-patterns` — see below.
- **The trailing `## References` index.** Folded into inline pointers at point of use.

## What stayed, and why

- **One merged anti-patterns table, 11 rows.** `Migration & gotchas` and `Anti-patterns` both carried the scope rename and, between them, five further traps. Merged into the single table the rubric requires, **preserving every distinct row and its reason** — host-fixed-by-login-path, the 60-day token lapse, Reel eligibility (9:16, 5–90s, H.264/HEVC), and the version-gated 2025–2026 metrics each became a row with its own *why it bites*. The scope row absorbed the full `instagram_basic` → `instagram_business_basic` mapping that only the migration bullet carried, so nothing was dropped to make the merge fit.
- **The six route-out rows.** Three of them (`shortform-strategy`, `video-shorts`, `api-connector-builder`) name siblings the 361-char description has no room for — genuine discriminative value, not restatement. All six targets verified to exist under `skills/`.
- **The bad/good publish pairs.** They teach poll-before-publish by demonstration rather than by shouting a rule.
- **The `status_code` decision table** (the flow genuinely branches), the **deprecated → replacement map with dates**, and the **OKF ingest front-matter block** that `scripts/verify.sh` greps.
- Both `references/` files stay linked — `publish-reel.md` at the end of the container dance, `insights-metrics.md` at the end of the deprecation trap.

**Before (723):** `… to publish a Reel from a video URL, pull per-media insights (…), check the 24-hour publish cap, or ingest Reel performance into 02-DOCS/wiki/shortform. Triggers: 'publish this Reel via the API', 'pull reach and average watch time for these media IDs', 'why does /insights?metric=plays return an error', 'my container is stuck IN_PROGRESS', … NOT TikTok publishing (that is tiktok-api), …`

**After (361):** `Use when wiring an agent into Instagram's Graph API: publishing a Reel through the create/poll/publish container dance, reading per-media insights, checking the 24h publish cap, or ingesting Reel metrics into 02-DOCS/wiki/shortform. NOT TikTok (that is `tiktok-api`), NOT YouTube (that is `youtube-api`), NOT cross-platform cadence (that is `social-publisher`).`

## Verification

- `bash scripts/eval-lint.sh` → `instagram-api  PASS (should_trigger=6, should_not_trigger=4, capability=1)`
- Frontmatter parses as YAML; `name` matches the directory, `origin: risco` present; all five `recommends` ids resolve to real skills.
- Every file under `skills/instagram-api/references/` is linked from the body.
- `manifest.json` untouched — the orchestrator regenerates it after merge.

### DIFF
--- a/skills/instagram-api/SKILL.md
+++ b/skills/instagram-api/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: instagram-api
-description: "Use when wiring an agent into Instagram's Graph API to publish a Reel from a video URL, pull per-media insights (reach, views, watch-time, saves, shares, total_interactions), check the 24-hour publish cap, or ingest Reel performance into 02-DOCS/wiki/shortform. Triggers: 'publish this Reel via the API', 'pull reach and average watch time for these media IDs', 'why does /insights?metric=plays return an error', 'my container is stuck IN_PROGRESS', 'how many posts left in the 24h publish window', 'publica este Reel por la API', 'ingesta el rendimiento dels Reels al wiki'. NOT TikTok publishing (that is tiktok-api), NOT a posting calendar/cadence (that is social-publisher), NOT YouTube uploads/stats (that is youtube-api)."
+description: "Use when wiring an agent into Instagram's Graph API: publishing a Reel through the create/poll/publish container dance, reading per-media insights, checking the 24h publish cap, or ingesting Reel metrics into 02-DOCS/wiki/shortform. NOT TikTok (that is `tiktok-api`), NOT YouTube (that is `youtube-api`), NOT cross-platform cadence (that is `social-publisher`)."
 tags: [instagram, graph-api, content-publishing, reels, insights, shortform]
 recommends: [tiktok-api, youtube-api, social-publisher, shortform-strategy, api-connector-builder]
 origin: risco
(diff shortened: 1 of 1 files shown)


ericrisco:

PR #153, opened 2026-07-29
refactor(harness): cut duplicated reference index, restated rules and description bloat

ericrisco:
Context-engineering pass on `harness` against `scripts/skill-rubric.md`. Format only — every claim, path, command and template that was in the skill is still in the skill.

| | Before | After |
|---|---|---|
| `description` | 470 chars | **344 chars** |
| `SKILL.md` | 359 lines / 28,039 B | **301 lines / 23,542 B** |
| eval-lint | — | **PASS** (should_trigger=7, should_not_trigger=6, capability=2) |

## Description

Kept the control-plane definition (the three layers) and the `NOT the bootstrap front door (that is init, which hands off here)` boundary — those are what let a reader pick this skill over its front door. Dropped the parenthetical examples of a non-code workspace and "Brownfield-first": characters paid for on every turn that never changed which skill gets chosen. No `Triggers:` list was present to remove.

## What went

- **The Architecture ASCII tree.** A second full index of `references/`, duplicating the tail References section line for line. Adding a reference meant editing two lists, and the two had already drifted — the tree listed `ingest-ignore-defaults.md`, the index listed `wiki-dashboard-template.html`, neither listed both.
- **The tail References section**, folded into point of use. Every `references/` file is still linked, now from the step that consumes it: `audit-report-template.md` + `examples/audit-example.md` at Phase 2, `agents-md-template.md` at APPLY 1, the seven wiki output templates at APPLY 8, `ingest-ignore-defaults.md` and `daily-curation-automation.md` in the 02-DOCS overview, `providers.yaml` in the red flag about a missing provider entry. Verified programmatically: 0 unlinked files under `references/`.
- **"When to use"** — restated the description, which is already in context on every turn. The two "Do NOT use" boundaries were substantive, so they moved up as an **Out of scope** line under Core principle, verbatim.
- **Invariants 1, 2, 3, 5, 6, 7 and 10.** Each was a word-for-word restatement of a phase step a few lines above (audit-before-action = Phase 2/3, explicit consent = Phase 3, no real `.env` = APPLY 3, merge-not-overwrite = APPLY 1, second destructive consent = APPLY 7, wiki-protocol authority = APPLY 8). A rule in an appendix is skimmed; a rule at the step is followed. The one piece of judgment that lived *only* in the appendix — never force-classify, and why — moved into APPLY 4 where classification happens.

## What stayed, on purpose

- **"Red flags — abort and re-plan".** Concrete failure modes with a named response, not rules already stated above — this is the skill's anti-pattern list, which the rubric requires.
- The five-phase protocol, the accepted consent forms, the `bash -n` syntax gate and the `python3` preflight **in full**, code fences included.
- The four remaining invariants — scope rules no single phase owns.
- "Keep CLAUDE.md lean", the Equip section, the Orientación block.
- **No result-envelope block added.** This skill never had one.

## Verification

- `bash scripts/eval-lint.sh` → `harness  PASS`
- All six `route_to` targets exist under `skills/`: `fastapi`, `init`, `deployment`, `postgresdb`, `nextjs`, plus one `none`. No stale "not in this catalog" claims in this skill's evals.
- `manifest.json` untouched; `npm test` / `npm run validate` not run (no `node_modules` in the worktree).
- One `why` string in `evals/cases.yaml` updated to cite "Out of scope" instead of the section name it replaced.

## For the orchestrator

`skills/harness/SKILL.md` has no `origin: risco` in its frontmatter — rubric deterministic gate 1. Pre-existing, not introduced here, and `skills/suggest/SKILL.md` is the only other skill in the catalog missing it (256/258 have it). The frontmatter schema does not currently declare or require `origin`, so `--validate` does not catch it. Left alone as a two-skill catalog fix rather than patched here.

### DIFF
--- a/skills/harness/SKILL.md
+++ b/skills/harness/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: harness
-description: "Use to control, govern or maintain a workspace's harness — software or non-code (a company, an ops desk, a knowledge vault). The harness is the control plane: the `01-TOOLS/` tooling layer, the `02-DOCS/` chaos→knowledge wiki, and the root Knowledge map. It audits, migrates legacy `XX-*` folders, scaffolds provider tooling, sweeps the inbox, and generates root CLAUDE.md/AGENTS.md. Brownfield-first. NOT the bootstrap front door (that is `init`, which hands off here)."
+description: "Use when governing a workspace's control plane, code or not — the `01-TOOLS/` tooling layer, the `02-DOCS/` chaos→knowledge wiki, the root Knowledge map. Audits it, migrates legacy `XX-*` folders, scaffolds provider tooling, sweeps the inbox, writes root CLAUDE.md/AGENTS.md. NOT the bootstrap front door (that is `init`, which hands off here)."
 tags: [harness, company, ops, docs, wiki, connect, tools, knowledge]
 recommends: [init]
 profiles: [minimal, core, full]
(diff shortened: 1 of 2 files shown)


ericrisco:
Superseded: harness is migrated and on main. Your trim landed via the wave-5 integration PR (#162); main additionally carries the `origin: risco` line added in #155, which is why this branch is not byte-identical to main.

Worth recording why this PR existed at all: the brief said the remote `skill/*` branches were the ledger of finished work, which stopped being true once integration PRs began merging with `--delete-branch`. A branch-based check re-offered `harness` after it was already done, and an agent migrated it a second time. The ledger is now the content — a migrated skill has no `Triggers:` list — fixed in #155.

PR #154, opened 2026-07-29
refactor(tauri): Claude-5-era context pass — description 782→339 chars

ericrisco:
Context-engineering pass on `skills/tauri/` against `scripts/skill-rubric.md`. No content rewrite: same versions, commands, config keys, and figures.

## Numbers

| | Before | After |
|---|---|---|
| `description` chars | 782 | **339** (target ≤350, hard limit 1024) |
| SKILL.md bytes | 11,470 | **10,650** |
| SKILL.md lines | 225 | **219** (ceiling 400) |

## What went, and why

- **The `Triggers:` keyword list** — eleven quoted phrases across English, Catalan and Spanish, more than half the description. The description is in context on every turn the skill is installed, invoked or not, so that was a per-turn cost for something the router already does by meaning. The rewritten description keeps what does the discriminating: the Rust-core-plus-OS-WebView value prop, the four surfaces it owns, and the `NOT a Chromium+Node shell needing Node APIs in a main process (that is electron)` boundary.
- **The trailing `## Pointers` index** — both `references/` files were already linked inline where they are used (`security.md` from the security section, `bundling-distribution.md` from the bundle section), and the sibling delegations were already stated in the intro and beside the CI matrix. Its two unique facts were folded inline rather than dropped: `secure-coding` moved into the intro's ownership sentence, and the `verify.sh` summary now sits directly under the anti-patterns table it checks.

## What stayed, and why

- **The anti-patterns table.** The rubric requires one, and these are concrete failure modes (blocking the IPC pool, unscoped `fs` permissions, a v1 `allowlist`, assuming bundled Chromium), not rules restated from above. It was already in `Anti-pattern | Why it's wrong | Do instead` form with no borrowed urgency framing, so no re-columning was needed.
- **The `std::sync::Mutex` → `tokio::sync::Mutex` bad/good pair.** It demonstrates the bad side, which the main worked example cannot show, and carries the "dropping the guard early creates a race" trap.
- **"Pick your starting shape".** A decision table where the flow genuinely branches (greenfield / existing web app / adding mobile / on v1).
- **Every rule bullet.** Each already sits next to the step it governs with a one-line *why*, which is the shape the rubric asks for — none were in an appendix rule bank.

No result-envelope block existed and none was added.

## Verification

- `bash scripts/eval-lint.sh` → `tauri PASS (should_trigger=7, should_not_trigger=5, capability=1)`
- Every `references/` file linked from the body: `security.md` ✓, `bundling-distribution.md` ✓
- Every `../<sibling>/SKILL.md` link resolves: `electron` ✓, `secure-coding` ✓
- Every eval `route_to` names a real skill: `electron`, `expo`, `rust`, `react`, `compose-multiplatform` — all present under `skills/`
- `scripts/verify.sh` is already mode `100755` and exits 0 on an empty target (gate 7)
- `manifest.json` untouched — it is derived and regenerated centrally

### DIFF
--- a/skills/tauri/SKILL.md
+++ b/skills/tauri/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: tauri
-description: "Use when building a lightweight cross-platform desktop (or v2 mobile) app with Tauri — a Rust core plus the OS-native WebView — wiring Rust commands and IPC, streaming data to the frontend, locking down the capabilities/permissions security model, or bundling and auto-updating tiny signed binaries. Triggers: 'build a Tauri app', 'expose a Rust command to my frontend', 'invoke from JS', 'tauri.conf.json', 'capabilities and permissions', 'stream download progress to the WebView', 'my command froze the UI', 'shrink my desktop binary vs Electron', 'set up the Tauri updater', 'crear una app d'escriptori lleugera amb Rust en lloc d'Electron', 'reducir el tamaño del binario de escritorio'. NOT a Chromium+Node desktop app that needs Node APIs in a main process (that is electron)."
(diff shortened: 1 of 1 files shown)


ericrisco:

PR #156, opened 2026-07-29
refactor(webhooks): Claude-5 rubric pass — description 1022→350, body 246→229 lines

ericrisco:
Context-engineering pass on `skills/webhooks/` against `scripts/skill-rubric.md`. No content rewrite — every technical claim, figure, command and code block is unchanged.

## Numbers

| | Before | After |
|---|---|---|
| `description` | 1022 chars | **350 chars** |
| `SKILL.md` | 13,869 bytes / 246 lines | **12,312 bytes / 229 lines** |

## Description

Removed the eight-phrase trilingual `Triggers:` list. It sat at 1022 chars — one char under the schema limit — and is in context on every turn the skill is installed, invoked or not, while the model routes by meaning anyway.

Kept the two boundaries that carry discriminative weight: `stripe` (the provider-specific scheme vs. the generic mechanics) and `api-connector-builder` (inbound receiver vs. the outbound client you write). `email-connector` and `automation-flows` were dropped from the description only — both still appear in the body's delegation list, and both still have `should_not_trigger` cases in `evals/cases.yaml`.

## What went

- **The duplicate cross-reference index.** The intro delegation list and the trailing `## Cross-references` section named the same five siblings twice. Merged into one list at the top; the one row the tail had that the intro did not (`secure-coding`) moved up, so nothing was lost.
- **An intro paragraph** that restated the five-stage pipeline three lines before the pipeline table itself.
- **"The handler must be fast"** as a standalone bullet — folded into the Fast-ack section, which owns the timeout → retry-storm → duplicate-processing chain. It was previously asserted in both places.
- **"non-negotiable"** from the pipeline heading and lead-in. The table's *"Why it precedes the next"* column already gives the reason each stage is ordered; the label added urgency, not information.

## What stayed, on purpose

- **The anti-patterns table** (9 rows). Required by dimension 7, and these are concrete failure modes with a *why it bites* column — not rules already stated above, so the hybrid rationalization-table treatment does not apply here. It was already in `Anti-pattern | Why it bites | Do instead` form.
- **The pipeline ordering table** and the **status-code contract table** — the flow genuinely branches at both.
- **Both bad/good code pairs** (parsed-body vs. raw-buffer HMAC; insert-then-process) — they teach the judgment by demonstration.
- **Every load-bearing fact**: the ~5-min tolerance window, the Standard Webhooks scheme and space-separated rotation signatures, `timingSafeEqual`/`compare_digest`, the dedupe-TTL-outlives-retry-window rule, the framework raw-body recipes.
- **No result-envelope block was added** — this skill never had one.

## Gates

- Both `references/` files (`signature-schemes.md`, `framework-raw-body.md`) are linked inline from the body.
- All `../<sibling>/SKILL.md` links resolve: `stripe`, `email-connector`, `secure-coding`. All six `recommends` ids exist.
- `scripts/verify.sh` was **already** executable (`100755`) and exits `0` on a clean tree — verified, no change needed.
- `manifest.json` untouched. `npm test` / `npm run validate` not run (no `node_modules` in the worktree) — for the global post-merge pass.

## Notes for the orchestrator

- No stale "not in this collection" claims in this skill — all five `route_to` targets in `evals/cases.yaml` (`stripe`, `email-connector`, `api-connector-builder`, `automation-flows`, `redis`) name real skills.
- The two overlapping structures here were cross-reference *lists*, not rule tables; handled by the same merge rule (one list, every distinct row preserved).

### DIFF
--- a/skills/webhooks/SKILL.md
+++ b/skills/webhooks/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: webhooks
-description: "Use when building the inbound side of a webhook — the HTTP endpoint that receives events a system pushes you, proves they are authentic, refuses to process the same one twice, and acks fast. Covers the mechanics every source shares: raw-body HMAC verification, ~5-min timestamp/replay window, idempotency on a stable event id, fast-ack-then-queue, retry/poison handling. Triggers: 'receive webhooks and process them safely', 'verify an incoming webhook signature', 'my signature verification fails even though the secret is correct', 'provider keeps retrying and we process the same event twice', 'handler times out under load, ack fast and queue', 'recibir webhooks entrantes', 'verificar la firma del webhook entrante', 'evitar procesar el mismo evento dos veces'. NOT the Stripe-Signature scheme (that is stripe), NOT email bounce/suppression hooks (that is email-connector), NOT outbound clients calling a third-party API (that is api-connector-builder), NOT what runs after the event lands (that is automation-flows)."
+description: "Use when building the receiving end of a webhook — a provider-agnostic endpoint that proves pushed events are authentic and never processes one twice: raw-body HMAC, ~5-min replay window, idempotency on the event id, fast-ack-then-queue. NOT the Stripe-Signature scheme (that is stripe), NOT outbound clients you call (that is api-connector-builder)."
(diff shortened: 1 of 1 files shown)


ericrisco:

PR #157, opened 2026-07-29
refactor(rag): cut the trigger list and merge the duplicated delegation map

ericrisco:
Context-engineering pass over `rag` against the current skill rubric. Format only — no recommendation, version, figure, threshold or command changed.

## Numbers

| | before | after |
|---|---|---|
| description | 967 chars | **346 chars** |
| SKILL.md | 226 lines / 13,249 bytes | **209 lines / 11,683 bytes** |
| eval-lint | PASS | **PASS** (7 / 5 / 1, unchanged) |

## Description

Dropped the eight-phrase `Triggers:` list spanning English, Spanish and Catalan. That list is in context on every turn the skill is installed, invoked or not, and the model matches by meaning without it. What remains is what discriminates `rag` from its nearest sibling: the pipeline it owns, the symptom that sends someone here (right document retrieved, answer still wrong), and the `NOT … (that is vector-db)` boundary the old description already carried.

> Use when building grounded Q&A over your own corpus — chunk, retrieve hybrid, rerank, ground, cite chunk ids, refuse when the sources fall short — or when the right document is retrieved but the answer is still wrong, invented, or unmeasured. NOT operating the store itself — collection schema, HNSW ef_search, quantization (that is `vector-db`).

## What went

- **The trailing "Wiring map".** It restated the stage table's *Hands off to* column plus the paragraph above it. Merged, not deleted: the reasons it added that the table lacked are now in the table cells (`document-processing` → "you assume text exists", `vector-db` → "you upsert, it owns the knobs", `embeddings-search` → "model, dims"), and its three entries that are **not** pipeline stages — `chatbot` ("it calls you, not the reverse"), `building-agents`, `structured-extraction` — moved into the paragraph under the table. Every sibling link and every distinct reason survives.
- **The tail `See references/…` index.** Both reference files were already linked inline at point of use, so it was a second copy of two links. The one detail only the tail carried — that `pipeline.md` holds a runnable retrieve → rerank → answer skeleton — moved into the inline mention, so the invariant holds: both files under `references/` are still linked from the body.
- **The boundary stated twice** in the opening two paragraphs; now once, keeping both tells (tuning HNSW / comparing embedding models) and both sibling links.
- **"three non-negotiable things" → "three things".** Two lines below, the body already says *why* it is absolute: "a grounding prompt without a refusal path is a bug". The shouting added nothing the reason did not.

## What stayed, deliberately

- **The anti-patterns table, all eight rows.** It was already framed `Anti-pattern | Why it breaks | Do instead` — no excuse/STOP wrapper to strip, and dimension 3 requires the table.
- **The stage table and the failure-symptom → metric table.** Both are genuine branches, not prose restated.
- **Every source, date and figure**: the ~35/49/67% contextual-retrieval reductions, the ~150 → ~20 funnel, Cohere Rerank 3.5, the RAGAS thresholds (faithfulness ≥ 0.90, context recall ≥ 0.85).
- **No result envelope was added** — `rag` never had one.

## Gates

- `bash scripts/eval-lint.sh` → `rag  PASS (should_trigger=7, should_not_trigger=5, capability=1)`; every `route_to` names a real skill (verified by hand — the linter does not check this).
- `scripts/verify.sh` was already mode `100755`; exits 0 (all 12 checks green).
- All seven siblings named in the body exist under `skills/`. Frontmatter parses; description 346 ≤ 1024.
- `manifest.json` untouched.

### DIFF
--- a/skills/rag/SKILL.md
+++ b/skills/rag/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: rag
-description: "Use when building grounded Q&A over your own documents — handbook, support tickets, PDFs, policies — and the answers must come only from the corpus with citations; when retrieval surfaces the right document but the answer is still wrong; when the model invents facts that are not in the sources; when you need to measure whether retrieval is any good; when the right document is found but the wrong passage is used. Triggers: 'build a bot that answers from our docs with citations', 'my RAG returns the right document but the answer is wrong', 'how do I stop the model hallucinating facts not in the corpus', 'context recall vs context precision', 'add contextual retrieval / prepend chunk context before embedding', 'hybrid search plus rerank pipeline', 'respuestas con citas sobre nuestros documentos', 'el bot s'inventa coses que no surten als documents'. NOT operating the vector store itself — collection schema, HNSW ef_search, quantization (that is vector-db)."
+description: "Use when building grounded Q&A over your own corpus — chunk, retrieve hybrid, rerank, ground, cite chunk ids, refuse when the sources fall short — or when the right document is retrieved but the answer is still wrong, invented, or unmeasured. NOT operating the store itself — collection schema, HNSW ef_search, quantization (that is `vector-db`)."
 tags: [rag, retrieval-augmented-generation, chunking, hybrid-search, reranking, grounding, citations, faithfulness, contextual-retrieval]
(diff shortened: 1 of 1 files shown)


ericrisco:

PR #161, opened 2026-07-29
refactor(gamedev-pathing): Claude-5 rubric migration

ericrisco:
Context-engineering pass on `gamedev-pathing` against `scripts/skill-rubric.md`. No content rewrite: recommendations, versions, APIs, commands and figures are unchanged.

## Numbers

| | Before | After |
|---|---|---|
| `description` | 975 chars | **349 chars** |
| `SKILL.md` | 16,420 bytes / 242 lines | **15,450 bytes / 234 lines** |
| eval-lint | — | **PASS** (should_trigger=7, should_not_trigger=5, capability=2) |

## Description

Before: an API catalogue plus `Triggers: "my enemies get stuck", "units clump together", "navmesh not updating", "agentes se atascan"`.

After:

> Use when making NPCs, enemies, or units navigate a level — grid vs waypoint vs navmesh, A*/JPS, navmesh baking and agent radius, off-mesh links, steering, RVO crowd avoidance, and flow fields in Godot 4.x, Unity, Unreal. NOT collision response or character controllers (that is gamedev-physics), NOT aggro or difficulty design (that is game-design).

The keyword list went because the model matches by meaning and this string is in context on every turn the skill is installed, invoked or not. Both `NOT` boundaries name real siblings (`skills/gamedev-physics`, `skills/game-design`) and are the nearest-neighbour discriminators the evals already test.

## What went

- **Two overlapping rule tables merged into one.** "Guardrails / gotchas" (6 bullets) and "Anti-patterns / rationalizations → STOP" (8 rows) covered overlapping ground — static bake vs runtime-spawned geometry was in both. They are now a single `Anti-pattern | Do instead` table of 13 rows: every distinct row and reason survives, the STOP / excuse framing does not.
- **Trailing "See Also — local references" index.** All three `references/` files are already pointed to inline where they are needed, so the tail was a duplicate link list. Verified each file is still reachable from the body.

## What stayed, deliberately

- **The version / API ban-list table.** Godot 4 renamed these and the old names fail silently — this is dimension-4 correctness, not padding.
- The two-layer path/steering model, the representation table, the search and navmesh essentials, all three per-engine mappings with their code blocks, the checklist, and the `02-DOCS` grounding pointer.

## Also

Linked `game-design` (Related skills) and the inline `gamedev-physics` mention — both were bare backticks naming siblings that exist.

`manifest.json` untouched; `npm test` / `npm run validate` not run in the worktree (no `node_modules`), per the migration brief.


ericrisco:

PR #163, opened 2026-07-29
refactor(gamedev-shaders): Claude-5-era rubric pass

ericrisco:
Claude-5-era rubric pass on `gamedev-shaders`. Format/context-engineering only — no
recommendation, version, command, or figure changed.

## Numbers

| | Before | After |
| --- | --- | --- |
| `description` | 937 chars | **353 chars** |
| `SKILL.md` | 15749 bytes / 253 lines | **14550 bytes / 232 lines** |
| eval-lint | PASS | **PASS** (7 trigger / 5 non-trigger / 1 capability) |

## Description

Dropped the `Triggers:` keyword list (six phrases across two languages) and the per-engine
feature inventory the body already carries. The new one leads with `Use when`, keeps the
engines that make it discriminative, and carries three explicit boundaries — including a new
one for `gamedev-shipping`, whose "strip shader variants in the build" case was a documented
near miss in the evals with no boundary in the description.

## What went

- **`When to use / When NOT to use`** — the *Use when* half restated the description; the
  *delegate* half was the same three routes as `Related skills`. Merged into `Related skills`,
  carrying over the nuances (a shader that *fakes* refraction is here, fluid simulation is not;
  this skill keeps the per-shader performance work).
- The intro's `Fires on …` sentence — description territory.
- The Unity section's restatement of the Version contract ban list (surface shaders,
  `UnityObjectToClipPos`) — it now points at the table instead of repeating it.
- The trailing `References` index — all three reference files are linked inline at point of use
  (`godot-shading-language.md`, `unity-and-unreal-shaders.md`, `effect-recipes.md`), so the
  every-reference-is-linked invariant holds.

## What stayed

- **`Guardrails / gotchas` → a real `Anti-patterns` table** (`Anti-pattern | Do instead`). Every
  distinct claim survives as a row; the three rows that only restated the Version contract table
  were folded into one row pointing at it. No new claims — each row is a failure mode already
  stated in the pipeline, performance, or post-processing sections.
- The **Version contract** table (the per-engine retired-API ban list, the heaviest correctness
  asset here), the coordinate-space and recipe tables, the Godot canvas_item/spatial/dissolve
  worked examples, `Performance`, and the `Checklist` — the same shape as the already-migrated
  `gamedev-pathing` sibling.
- No result-envelope block existed; none was invented.

## Evals

Two `why` strings cited trigger phrases that no longer exist in the description ("an explicit
trigger", "the description carries the Spanish …"); reworded to the semantic reason. Prompts,
`route_to` values, and the capability rubric are untouched. All five `route_to` targets are real
skills (`godot`, `unity`, `gamedev-physics`, `gamedev-shipping`, `sql`) — no stale
"not in this catalog" claims in this one.


ericrisco:

PR #165, opened 2026-07-29
refactor(notion-connector): Claude-5-era rubric pass — 865→347 char description

ericrisco:
Claude-5-era rubric pass on `notion-connector`. Format/context-engineering only — no recommendation, version, command, endpoint or figure was changed.

| | Before | After |
|---|---|---|
| `description` | 865 chars | **347 chars** |
| body | 10800 bytes / 234 lines | **9797 bytes / 221 lines** |

## What went

**The `Triggers:` list.** Seven phrases across English, Spanish and Catalan, in context on every turn this skill is installed whether or not it fires. The model matches by meaning; the list was pure per-turn cost. The two `NOT` boundaries stay and got tighter — `api-connector-builder` (generic REST wiring) and `webhooks` (inbound events), both verified present under `skills/`.

**`## When to use / when NOT to use`.** Its `Use when:` paragraph was the description restated one screen later. Deleted. The routing table beneath it is a real branch, not a restatement, so it survives intact as `## Route elsewhere` — all five rows, all five sibling links.

**`## The one rule` as a standalone block.** It asserted the two things that matter, and then setup step 4 asserted the pinning half again and the data-source section asserted the resolution half again. That is the repeat-across-layers pattern the rubric names. Both halves moved next to the step they govern rather than being deleted:

- the version facts — current major `2025-09-03` is the SDK default, latest is `2026-03-11`, behavior differs across versions so pin per client or per request, an unpinned client drifts when the default moves — now sit in setup step 4, at the line where the client is actually constructed;
- *a database is a container of one or more data sources, not a queryable table; resolve it before you query or last year's code 404s silently* now opens `## The database → data source model`, the section that explains it.

No claim was dropped in the move.

## What stayed, deliberately

- **The 10-row anti-patterns table, including its `Why it bites` column.** Not a rationalization table wearing a different hat: every row names a concrete failure mode and the symptom it produces (404 on the retired `databases/:id/query` path, silently dropped rows past 100, duplicate rows from a blind `pages.create`, `archived` vs `in_trash`). The symptom column is what makes it diagnostic rather than a restatement of the steps.
- **Every decision table.** `database_id` → `data_source_id` resolution, the property write shapes, and the version-migration diffs all branch genuinely.
- **No result-envelope block was added.** This skill never had one; inventing it would be a content change.
- All three `references/` files remain linked inline from the body (`property-shapes.md` ×2, `api-versions.md` ×2, `sync-patterns.md`), so nothing in the package went dead.

`scripts/verify.sh` was already executable and untouched. `manifest.json` deliberately not edited — it is derived and regenerated centrally.

### DIFF
--- a/skills/notion-connector/SKILL.md
+++ b/skills/notion-connector/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: notion-connector
-description: "Use when wiring an app, cron job, or service to Notion as an ops backend over the official HTTP API — pushing rows into a Notion database, mirroring a database into your app, syncing both ways without duplicates, building or reading page block content, or fixing an integration that broke after the 2025-09-03 data-source split. Triggers: 'store our tasks in a Notion database from code', 'use Notion as a CRM/content calendar', 'sync our CRM into Notion nightly without duplicates', 'why does POST /v1/databases/:id/query return 404 now', 'integration broke after a teammate added a second data source', 'guardar registros en una base de datos de Notion desde código', 'desar files a una base de dades de Notion'. NOT a generic call-any-REST-API integration (that is api-connector-builder), NOT receiving/verifying inbound Notion webhook events (that is webhooks)."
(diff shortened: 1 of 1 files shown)


ericrisco:

PR #166, opened 2026-07-29
refactor(n8n): trim description to 355 chars, fold checklist and gotchas into one anti-patterns table

ericrisco:
Context-engineering pass on `n8n` against `scripts/skill-rubric.md`. Format only — no claim, version, endpoint, or command changed.

## Numbers

| | Before | After |
|---|---|---|
| `description` | 995 chars | **355** |
| body | 12242 bytes / 206 lines | **10329 bytes / 177 lines** |

## Description

Dropped the nine-phrase `Triggers:` list (EN + ES). It is in context on every turn the skill is installed, invoked or not, and the model matches these by meaning anyway.

Kept the two boundaries that actually discriminate — `automation-flows` (design and the importable JSON) and `make` / `zapier` (other platforms) — with the siblings in backticks. The `api-connector-builder` and `webhooks` boundaries moved into the body's route table, so they are only read when the skill fires. Both siblings still appear in `recommends`.

## What went

- **The 8-item Checklist.** Every item restated a step already above it: the `.env` keys, the payload shape, validate-before-create, capture the returned id, test, activate via the endpoint, export before delete, `DISABLED_TOOLS`. The two absolutes moved to the steps they govern — `tags` is now named next to `active` in step 4, and "export before `PUT`" sits with the delete commands.
- **The "Related skills" bullet list**, re-columned as a compact `If the ask is | Skill` table. Same five routes, same "neither Zapier nor Power Automate can create flows via API" note.
- **Two within-page duplicates:** "the API surface is the same on self-host and cloud" (already a row in the REST-vs-MCP table) and the second telling of "cloud has no Swagger" (kept where the fact already lived, in Version reality).
- **The trailing References index.** `auth-and-selfhost.md` and `n8n-mcp-tools.md` were already linked at point of use; `rest-api-cheatsheet.md` now is too, from the lifecycle header. All three references remain linked — an unlinked reference never loads.

## What stayed, deliberately

- **All seven "Honesty / gotchas" facts**, re-columned as `Anti-pattern | Why it bites | Do instead` — the shape `automation-flows`, `api-connector-builder` and `cloudflare` already use. These are concrete failure modes (read-only `active`, no-input `execute`, full-replace `PUT`, a missing `typeVersion`, unreadable credential secrets, API activation skipping UI validations), not rules restated from above.
- The REST-vs-MCP decision table (the flow genuinely branches on who is driving), the full curl/MCP lifecycle, the version reality (PR #20234, `/api/v1/docs` self-host only), the `.env` block, and Governance.

## Checks

- `bash scripts/eval-lint.sh` → `n8n PASS`
- Frontmatter parses as YAML; `description` 355 chars (well under the 1024 schema limit)
- Every `references/` file linked from the body; `../automation-flows/references/n8n-workflow-json.md` still resolves

### DIFF
--- a/skills/n8n/SKILL.md
+++ b/skills/n8n/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: n8n
-description: "Use when operating a live n8n instance programmatically — driving its public REST API (`{host}/api/v1`, `X-N8N-API-KEY`) or the n8n-mcp server to create, validate, test, activate/deactivate, list, update, or delete workflows and read executions, on self-host or `*.app.n8n.cloud`. Triggers: 'deploy this workflow to my n8n over the API', 'push my workflow JSON to n8n with curl', 'I set active:true in the create call but it stays inactive', 'connect n8n-mcp / n8n_create_workflow to Claude', 'validate and autofix my n8n workflow', 'run an n8n workflow by id from a script', 'delete a workflow via the API', 'automatiza n8n por API', 'desplega el workflow a n8n.cloud'. NOT designing the flow's trigger-step-error logic or hand-building the importable workflow JSON (that is automation-flows), NOT operating Make / Zapier / Power Automate (their own skills), NOT a typed API client with auth+pagination+retry (api-connector-builder), NOT the inbound webhook receiver in your own app (webhooks)."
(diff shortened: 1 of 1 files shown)


ericrisco:

PR #167, opened 2026-07-29
refactor(knowledge-ops): Claude-5-era description and body pass

ericrisco:
Context-engineering pass on `knowledge-ops` against `scripts/skill-rubric.md`. Substance unchanged — no recommendation, threshold, path or command was touched.

| | Before | After |
|---|---|---|
| `description` | 873 chars | **344 chars** |
| `SKILL.md` | 11,640 bytes (125 lines) | **10,729 bytes (122 lines)** |

## What went

- **The `Triggers:` list** — eight phrases across Spanish and English, paid for on every turn the skill is installed whether or not it fires. The description is now what-it-does / when / boundary, naming `harness` explicitly as the sibling it has to be told apart from.
- **The reserved-file rule, stated five times.** `log.md` is newest-first, `gaps.md` is append-only, past entries are immutable — this appeared under the four-operations table, twice in Prune, and as two separate safety rails (#3 and #4, which were the same rule split in half). Merged into one rail that preserves every clause of all five statements; the other sites now point to it.
- **The duplicated harness/knowledge-ops distinction.** The intro paragraph and the closing "sharp line" paragraph made the same discrimination. Folded into the intro; the trailing copy is gone (the hand-off table's harness row still says "it owns the engine").
- **A duplicate `references/gardening-playbook.md` pointer** two lines apart in Prune, and the `(inherited from the protocol)` heading suffix that the very next line restates.

## What stayed, deliberately

- **The anti-patterns table, all eight rows.** It names concrete failure modes (`rm`-ing a superseded article, auto-picking a conflict winner, re-running `harness` to fix a link, splitting on length alone), which is what dimension 3 requires — not a rationalization bank.
- The four-operations decision table, the capture altitude ladder, the bad/good title and nesting pairs, and the full content of every safety rail.
- The pre-existing hand-off route table. `knowledge-ops` has no result-envelope block and none was invented.

## Fixes carried along

- Three siblings were named in bare backticks in the hand-off table — `decision-records`, `codebase-onboarding`, `research-ops` — now linked as `../<id>/SKILL.md` like the rest. All six route targets verified present under `skills/`.
- `evals/cases.yaml` needed no change: every `route_to` names a real skill, and no `why` string quoted a deleted trigger phrase.
- `references/gardening-playbook.md` is still linked from the body (three inline mentions), so nothing in the package became unreachable.

`manifest.json` untouched.

### DIFF
--- a/skills/knowledge-ops/SKILL.md
+++ b/skills/knowledge-ops/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: knowledge-ops
-description: "Use when an already-running 02-DOCS/ wiki needs day-to-day gardening — too many half-finished articles, a 600-line page that covers three unrelated topics, a note with no obvious home, orphan pages nothing links to, stale answers and superseded articles piling up, or two articles that contradict each other. This is the judgment layer over harness's automation: decide what is worth capturing (default: nothing), where it goes, how to title/split/merge it, build the See Also web, and prune to _archive (never delete). Triggers: 'limpia mi wiki', 'organiza mis notas', 'esto dónde va en la wiki', 'ordena la base de conocimiento', 'split this bloated article', 'where should this note live', 'scores.json says these three are orphans — fix the cross-links', 'prune the stale and superseded stuff'. NOT building, bootstrapping or sweeping the wiki engine (that is harness)."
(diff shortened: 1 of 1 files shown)


ericrisco:

PR #168, opened 2026-07-29
refactor(gcp-essentials): cut the trigger list and the tail reference index

ericrisco:
Per-skill context pass on `gcp-essentials` under the current rubric. Format only — no recommendation, version, flag, command or figure changed.

| | Before | After |
|---|---|---|
| `description` | 588 chars | **362 chars** |
| body | 15222 bytes / 328 lines | **14930 bytes / 325 lines** |

## What went

- **The `Triggers:` list** — nine phrases across English and Spanish (`'deploy to Cloud Run'`, `'crear bucket en GCP'`, `'service account sin claves JSON'`, …). It is paid for on every turn the skill is installed, invoked or not, and the model matches Cloud Run / Cloud SQL / bucket intent by meaning without it.
- **The trailing `## References` index.** All three reference files were already linked at their point of use, so the index was a second copy of the same three links. The detail it carried was preserved by moving it into those inline mentions rather than deleted: the predefined-role catalog onto `iam-and-auth`, Direct VPC egress vs legacy connectors and pooling onto `networking-and-sql`, and the runbook list onto `deploy-recipes` at the end of "Wire it together", where someone wanting copy-paste actually is. All three files remain linked.

## What changed shape

- The boundary went from one bare `NOT AWS (that is aws-essentials)` to three backticked siblings — `aws-essentials`, `deployment`, `postgresdb` — the ones this skill is genuinely confused with. Discrimination over coverage: Cloud SQL operations here, Postgres query tuning there.
- Siblings that were sitting in bare backticks (`aws-essentials`, `docker`, `github-actions`, `monitoring`, `backups`) are now links, matching the ones in the same block that already were.

## What stayed, deliberately

- **The anti-patterns table, all ten rows**, already in `Bad | Good | Why` form. Concrete failure modes (default compute SA, JSON keys, `allUsers` buckets, secrets in `--set-env-vars`), not rules restated from above.
- **Both IAM hard rules.** They sit next to the Cloud Run step they govern and each states its own consequence — an RCE becoming project takeover, a long-lived leaked credential — rather than an appendix shouting NON-NEGOTIABLE.
- **The member-type and role-kind tables.** Distinct content, not overlapping, so nothing to merge.
- **Section 5, "Wire it together."** It overlaps `references/deploy-recipes.md`, but de-duplicating body against references is a content change, not a format one.

## Not touched

- `evals/cases.yaml` — no `why` string quoted a deleted trigger phrase, and all four `route_to` ids (`aws-essentials`, `postgresdb`, `deployment`, `secure-coding`) name real skills.
- `scripts/verify.sh` — already executable.
- `manifest.json` — derived; regenerated centrally after merge.

### DIFF
--- a/skills/gcp-essentials/SKILL.md
+++ b/skills/gcp-essentials/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: gcp-essentials
-description: "Use when running a small product on the core of Google Cloud with the gcloud CLI: creating a project, deploying a container to Cloud Run, standing up a Cloud Storage bucket, a managed Cloud SQL Postgres/MySQL, and wiring them together with least-privilege IAM. Triggers: 'deploy to Cloud Run', 'gcloud run deploy', 'set up a GCP project', 'crear bucket en GCP', 'desplegar a Cloud Run', 'service account sin claves JSON', 'why is my Cloud Run running as the default service account', 'connect Cloud Run to Cloud SQL', 'turn off the bucket public access'. NOT AWS (that is aws-essentials)."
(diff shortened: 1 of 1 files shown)


ericrisco:

PR #169, opened 2026-07-29
refactor(hetzner): Claude-5 rubric pass — 684→365 char description, posture block out

ericrisco:
Claude-5-era context pass on `hetzner`, per `scripts/skill-migration-brief.md`.

## Numbers

| | Before | After |
|---|---|---|
| `description` | 684 chars | **365 chars** (-47%) |
| `SKILL.md` | 11,385 bytes / 221 lines | **10,544 bytes / 203 lines** |

## What went, and why

**Description.** The eight-phrase `Triggers:` list went — English plus the
Spanish/Catalan duplicates. That list sat in context on every turn the skill was
installed, invoked or not, to do a job the model already does by meaning. The
boundary was a single `NOT coolify`; it is now the three siblings a reader
actually confuses this with — `coolify` (app deploys), `docker`
(Dockerfiles/images), `digitalocean` (another provider's VPS) — so the
description discriminates rather than covers. All three exist under `skills/`
and all three are the `route_to` targets of existing negative eval cases.

**Body.**

- The **"Operating posture"** bullet block. Every one of its three rules was
  restated at the step it governs: reproducibility by the "never harden it later"
  rule in *Provision reproducibly*, the edge-first firewall order by the entire
  *Two firewalls, in order* section, and the no-SLA/no-managed-DB honesty by
  *Day-2*. Two facts in it were **not** duplicated, so both survived: the "a box
  you can't recreate from a file is a pet" line moved into the intro, and
  "phone support only for dedicated-server customers" folded into the Day-2
  no-SLA bullet.
- The **bad/good sshd diff**. It repeated the two bullets directly above it, which
  in turn repeat the cloud-init `sed` lines. `PubkeyAuthentication yes` was its
  only unique content and moved into the key-only bullet.
- Two lines of hedging in the Docker/Coolify handoff, now one direct route.

## What stayed, and why

- **The anti-patterns table, all eight rows, columns untouched.** The rubric
  requires one, and this is the real thing rather than a rationalization bank: it
  names concrete failure modes (IPv6-only breaking `docker pull`, US/Singapore
  egress caps, the `--datacenter` removal) that appear nowhere else in imperative
  form.
- **The plan/location decision table**, the `Rules:` bullets with their `Why:`
  clauses, the host-ready checklist, and the `Verify` block. The flow genuinely
  branches there and the reasoning is the actionable part.
- **Every load-bearing fact verbatim** — hcloud v1.65.0 (2026-05-21), the
  2026-04-01 price adjustment, the 2026-07-01 `datacenter` removal, the €/mo
  bands, the 20 TB EU egress. This was a format pass; no recommendation, version,
  command, or figure changed.
- No result-envelope block existed, so none was invented.

## Gates

- `bash scripts/eval-lint.sh | grep '^hetzner'` → **PASS**.
- Frontmatter parses as single-line quoted YAML; description 365 ≤ 1024.
- Both `references/` files still linked inline (`plans-and-locations.md`,
  `cloud-init.md`) — the invariant holds, no tail index was needed.
- Sibling mentions in the body are now real `../<id>/SKILL.md` links
  (`coolify`, `docker`, `backups`, `domains-dns`, `monitoring`); all resolve.
- `scripts/verify.sh` already executable; re-confirmed read-only and exit 0 on an
  empty target.
- `manifest.json` untouched — it is derived and regenerated centrally.

## Notes for the orchestrator

Nothing stale found in `evals/cases.yaml`: all five `route_to` values name real
skills, and no `why` string quoted a deleted trigger phrase. The Spanish
`should_trigger` prompt is a genuine case about the prompt language, not an echo
of the removed keyword list, so it stays.

### DIFF
--- a/skills/hetzner/SKILL.md
+++ b/skills/hetzner/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: hetzner
-description: "Use when provisioning or hardening a Hetzner Cloud VPS to host real apps: picking a plan/location (CX/CPX/CAX/CCX), bringing a box up reproducibly with the hcloud CLI + cloud-init, locking down SSH, and wiring the Hetzner Cloud Firewall before handing off a Docker/Coolify-ready host. Triggers: 'set up a Hetzner server', 'provision a Hetzner VPS', 'harden SSH on my box', 'which Hetzner plan for a small API', 'configure the Hetzner Cloud Firewall', the non-obvious 'my ipv6-only box is cheaper but docker pull can't reach the registry', and the Spanish/Catalan 'monta un servidor en Hetzner', 'configura el firewall de Hetzner'. NOT deploying apps through Coolify (that is coolify)."
+description: "Use when provisioning or hardening a Hetzner Cloud VPS: picking the plan/location (CX/CPX/CAX/CCX), bringing it up reproducibly with hcloud + cloud-init, locking down SSH, and wiring the Cloud Firewall to hand off a Docker-ready host. NOT app deploys on Coolify (that is `coolify`), NOT Dockerfiles (that is `docker`), NOT another provider (that is `digitalocean`)."
 tags: [hetzner, vps, cloud-firewall, hcloud, ssh-hardening, cloud-init, self-hosting]
 recommends: [coolify, docker, secure-coding, domains-dns, backups, monitoring]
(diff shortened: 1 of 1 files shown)


ericrisco:

PR #170, opened 2026-07-29
refactor(planetscale): Claude-5-era rubric pass

ericrisco:
Context-engineering pass on `planetscale` against `scripts/skill-rubric.md`. Format only — no
recommendation, version, command, or figure changed.

| | Before | After |
| --- | --- | --- |
| `description` chars | 913 | **353** |
| SKILL.md bytes | 13330 | **11302** |
| SKILL.md lines | 223 | **198** |

## Description

Dropped the ten-phrase `Triggers:` list (EN + ES). It was in context on every turn the skill is
installed, invoked or not, and the body already routes by meaning. What is left is the rubric shape:
what it does · when · an explicit NOT boundary.

The old boundary read `NOT raw MySQL query tuning … (that is postgresdb's MySQL-engine analogue /
the mysql sibling)` — a hedge from when `mysql` had not shipped. It has, so the boundary now names
`mysql` and `db-migrations` directly. Both verified present under `skills/`.

## What went

- **"When to use" bullets** — a restatement of the description, paid for twice.
- **The trailing routing paragraph**, which repeated the "would this answer be identical on RDS
  MySQL?" heuristic already stated above it.
- **The "Deep dives" index line.** Both `references/` files are linked inline at the point of use
  (`deploy-requests` under the workflow, `no-foreign-keys` under the FK section), so nothing became
  unreachable — the package invariant holds.
- **The "closest built sibling" hedge** around MySQL-engine work, now that `mysql` exists.

## What stayed, on purpose

- **All 7 anti-pattern rows.** Concrete failure modes (`ALTER` on `main`, direct `mysql://` from
  Vercel Edge, inventing `@planetscale/serverless`), not rules restated from the flow above them.
- **Both FK-strategy rows**, moved out of a trailing `## Decision:` appendix and placed next to the
  no-FK section they govern. A rule in context is followed; a rule in an appendix is skimmed.
- **All of the "When NOT to use" routing substance**, re-columned into a single table so every
  sibling and its rationale survives — including the PlanetScale-sells-Postgres-too caveat and the
  "this skill owns only `relationMode` and the serverless adapter" line.
- The mental-model table, the numbered `pscale` lifecycle, both Bad→Good fences, the
  runtime/connection table, and the ~30-minute revert-window carve-outs — verbatim.

No result-envelope block existed here, and none was invented.

## Hygiene

- Siblings that sat in bare backticks are now links: `mysql`, `db-migrations`, `neon`, `supabase`,
  `prisma-orm`, `github-actions`.
- Fixed two links in `references/no-foreign-keys.md` that used `../<sibling>/SKILL.md` from inside
  `references/` and so resolved to nothing; they are `../../` now.
- `scripts/verify.sh` was already executable and still exits 0 (its four content gates pass against
  the trimmed body).
- `evals/cases.yaml` untouched — none of its `why` strings quoted the deleted trigger phrases, and
  all five `route_to` targets are real skills.
- `manifest.json` not touched.

## Verification

```
$ bash scripts/eval-lint.sh | grep '^planetscale'
planetscale            PASS (should_trigger=7, should_not_trigger=5, capability=1)

$ bash skills/planetscale/scripts/verify.sh; echo $?
[ok] verify.sh passed
0
```

Frontmatter parses as YAML, `description` is a single-line quoted scalar at 353 chars, `name`
matches the directory, `origin: risco` intact. `npm test` / `npm run validate` not run — no
`node_modules` in the worktree, per the brief.

### DIFF
--- a/skills/planetscale/SKILL.md
+++ b/skills/planetscale/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: planetscale
-description: "Use when operating PlanetScale serverless MySQL (Vitess) — setting up a database/branch and wiring the pscale CLI, shipping a schema change through branches and deploy requests, debugging a stuck/queued deploy or a cutover that won't finish, designing tables without foreign keys, or picking the serverless HTTP driver for an edge runtime. Triggers: 'schema change without downtime on PlanetScale', 'my migration is stuck in the deploy queue', 'the cutover hung', 'can I still revert 40 minutes after deploy', 'the 30-minute revert window', 'PlanetScale won't let me add a foreign key', 'relationMode prisma', 'MySQL connection times out on Vercel Edge', 'cambio de esquema sin downtime', 'no me deja usar foreign keys'. NOT raw MySQL query tuning, EXPLAIN, or index choice (that is postgresdb's MySQL-engine analogue / the mysql sibling), and NOT generic expand-contract migration theory (that is db-migrations)."
+description: "Use when operating PlanetScale (Vitess serverless MySQL) — branches and the `pscale` CLI, schema changes through deploy requests (Online DDL cutover, ~30-min revert window), tables without foreign keys, or the edge HTTP driver. NOT MySQL engine work like EXPLAIN or index choice (that is `mysql`), NOT generic migration theory (that is `db-migrations`)."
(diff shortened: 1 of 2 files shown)


ericrisco:

PR #172, opened 2026-07-29
refactor(linkedin-carousels): 902 -> 352 char description, de-duplicated boundary prose

ericrisco:
Context-engineering pass on `linkedin-carousels`, per `scripts/skill-migration-brief.md`.

## Numbers

| | Before | After |
|---|---|---|
| `description` | 902 chars | **352 chars** (-61%) |
| body | 14209 bytes / 239 lines | **13104 bytes / 229 lines** |

The description is the part that matters: it sits in context on every turn the skill is installed, invoked or not.

## What went

- **The `Triggers:` list** — 8 phrases across English, Spanish and Catalan, roughly 430 of the 902 chars. The model matches by meaning; enumerating `'turn this into a LinkedIn carousel'` next to a body that already routes semantically is pure per-turn cost.
- **Decorative adjectives on the deliverable** — "scroll-stopping" cover hook, an arc "that pulls swipe-to-swipe", "real visual hierarchy", and the "currently the highest-reach native LinkedIn format" claim. The body's swipe-economics section states that properly, with its figure and its reason.
- **The boundary paragraph in the body** — it named four of the nine routing targets already in the Handoffs table directly below it. Two lists of the same decisions drift apart; the table is now the single authority and keeps every row.
- **A pointer to `evals/`** in the Verify section. The reader never loads evals, so the line was dead weight.
- Minor reflow of four paragraphs that were split for no structural reason (Ground first, Slide count, the export rationale, the caption handoff).

## What stayed, deliberately

- **The full 10-row anti-patterns table, all three columns.** This is not a rationalization bank — every row names a concrete failure mode with a real consequence (Canva silently substituting fonts at PDF export, bleed marks adding black bars, a 16:9 canvas reading as a talk deck). Dimension 3 requires the table and these rows are the skill's hard-won detail.
- **All three worked Bad→Good ASCII slide sketches.** They teach visual hierarchy by demonstration; prose cannot replace a sketch of a crammed slide next to a clean one.
- **The 5-row narrative-arc decision table** — the flow genuinely branches on the angle.
- **"What you produce"** — an output contract that already existed. Per the brief, existing contracts stay and absent ones are not invented.
- **The runnable `scripts/verify.sh` fence.** Actionability over byte count; it is a real command a practitioner runs.
- **Every load-bearing figure, untouched**: 6.6% average engagement, the 6–12 band with ~8 as the sweet spot, 7–8 reported peaks, 1080×1350 / 1080×1080, ~880×880 safe area, ~80 px padding, ≥28 px type floor, ~100-300 MB / 300 pp cap, < 10 MB target, ~7% completion drop per extra field. This was a format pass, not a content rewrite.

## Incidental fix

`linkedin-outreach` was the only sibling in the Handoffs table left as bare backticks while the other eight were real relative links. Linked it — the skill exists at `skills/linkedin-outreach/`.

## Verification

- `bash scripts/eval-lint.sh` → `linkedin-carousels  PASS`
- `description` 352 chars, parses as single-line quoted YAML, well under the 1024 schema limit.
- All 9 `../<sibling>/SKILL.md` links resolve to real skills.
- The one file under `references/` (`carousel-patterns.md`) is still linked from the body, three times, inline at point of use.
- `scripts/verify.sh` is executable and exits 0 on an empty target.
- `manifest.json` untouched — it is derived and the orchestrator regenerates it.

### DIFF
--- a/skills/linkedin-carousels/SKILL.md
+++ b/skills/linkedin-carousels/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: linkedin-carousels
-description: "Use when designing a LinkedIn carousel / document post — the swipeable multi-page PDF that is currently the highest-reach native LinkedIn format — and you need the slide-by-slide build: a scroll-stopping cover hook, a narrative arc that pulls swipe-to-swipe, one focused idea per slide with real visual hierarchy, a payoff + CTA closer, and a PDF export spec. Triggers: 'turn this into a LinkedIn carousel', 'design a swipeable PDF document post', 'my carousel cover gets no swipes fix the hook', 'how many slides and what goes on each', 'restructure one-idea-per-slide', 'what size to export so it isnt cropped on mobile', 'convierte este post en un carrusel de LinkedIn', 'dissenya'm un carrusel de LinkedIn amb portada i CTA'. NOT the account/format plan (that is linkedin-strategy), NOT a text or single-image feed post (that is linkedin-content), NOT a spoken-to talk deck (that is presentations)."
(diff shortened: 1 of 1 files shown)


ericrisco:

PR #173, opened 2026-07-29
refactor(lead-gen): Claude-5-era description and body pass

ericrisco:
Context-engineering pass on `lead-gen` against `scripts/skill-rubric.md`. Format only — no recommendations, versions, figures, commands or claims were changed.

| Metric | Before | After |
|---|---|---|
| `description` chars | 670 | 333 |
| SKILL.md bytes | 11350 | 10697 |
| SKILL.md lines | 155 | 151 |
| `scripts/eval-lint.sh` | — | **PASS** (should_trigger=6, should_not_trigger=5, capability=1) |

## What went

- **The `Triggers:` list** — nine phrases across English, Spanish and Catalan. That description is in context on every turn the skill is installed, invoked or not, so a keyword list the model does not need is pure per-turn cost. Half the budget of the field went to it.
- **A restated identity paragraph.** "The identity of this skill is **the list and its scoring model**…" repeated the italic intro one line above it; its one non-duplicated clause (roster + rationale, handoff is the finish line) was folded into the intro.
- **A trailing routing restatement.** "When the deliverable would be a message, a pipeline stage, or a market sizing — route." said exactly what the three handoff bullets immediately above it already say.

The boundary against `cold-outreach` and `sales-pipeline` was stated five times — description, intro, anti-patterns row, handoff bullets, closing line. Four remain, each doing different work.

## What stayed, deliberately

- **The anti-patterns table.** Already in the right shape (`Anti-pattern | Why it fails | Do instead`), concrete failure modes with no borrowed urgency framing — the rubric requires one.
- **The branching tables**: BANT/CHAMP/MEDDIC by deal size, the provider/waterfall comparison, and tier→SLA. These are real decision points, not padding.
- **The bad/good ICP pair**, which teaches falsifiability by demonstration rather than by rule.
- **The per-phase `*Why:** notes.** These are rules sitting next to the step they govern instead of in an appendix — the shape the brief asks for, not the shape it removes.
- **The compliance gate** (GDPR LIA / CAN-SPAM) with its dated sources, and the `scripts/verify.sh` pointer.
- Both `references/` files remain linked inline at point of use (`data-sources.md` in phase 2, `scoring-model.md` in phase 3).

No result-envelope block existed and none was invented.

## Evals

Two `why` fields framed the Spanish and Catalan cases as "triggers"; reworded to "…-language phrasing", since the keyword list they pointed at is gone. The cases themselves are unchanged. All five `route_to` targets verified to exist under `skills/`: `cold-outreach`, `sales-pipeline`, `market-research`, `data-scraper`, `email-deliverability`.

## Notes

`manifest.json` deliberately untouched — it is derived and regenerated centrally. `scripts/verify.sh` was already executable. No stray markup fragments, no overlapping rule tables, no stale "not in this catalog" claims, and no bare-backtick sibling references needing links.

### DIFF
--- a/skills/lead-gen/SKILL.md
+++ b/skills/lead-gen/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: lead-gen
-description: "Use when you need to find and qualify prospects, build a target list, or score and prioritize leads before anyone reaches out — defining an ICP, sourcing named accounts/contacts from Apollo/ZoomInfo/Clay, deduping against the CRM, and ranking with a fit+intent+engagement model. Triggers: 'build me a prospect list', 'who should we sell to', 'score these leads', 'which accounts do we call first', 'tighten our ICP', 'is this cold list GDPR-safe', 'necesito una lista de clientes potenciales', 'califica estos leads', 'a qui truquem primer'. NOT writing or sending the outreach (that is cold-outreach), NOT tracking the deal after first contact (that is sales-pipeline)."
+description: "Use when building and qualifying a prospect list before anyone reaches out — a falsifiable ICP, named accounts/contacts from Apollo/ZoomInfo/Clay, deduped against the CRM, tiered by fit+intent+engagement. NOT writing or sending the outreach (that is cold-outreach), NOT tracking the deal after first contact (that is sales-pipeline)."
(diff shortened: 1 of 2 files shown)


ericrisco:

PR #174, opened 2026-07-29
refactor(llm-pipeline): Claude-5-era rubric pass

ericrisco:
## What

Context-engineering pass on `llm-pipeline` per `scripts/skill-migration-brief.md`. No content rewrite: no recommendation, version, price, command or figure changed.

| | Before | After |
|---|---|---|
| `description` chars | 690 | 347 |
| body bytes | 12782 | 12023 |
| body lines | 155 | 151 |

## Description

Removed the nine-phrase `Triggers:` list (EN/ES/CA). It is in context on every turn the skill is installed, invoked or not, and the model matches by meaning anyway. Rewrote to what-it-does / when / boundary, and made the boundary discriminate against **two** real siblings instead of one:

> Use when wiring several LLM calls into a production flow: typed contracts between steps, a router/gateway so 429s, timeouts and outages fail over instead of taking you down, and cost control via caching, model tiers and abort caps. NOT single-prompt wording (that is `prompt-engineering`), NOT a model-driven tool loop (that is `building-agents`).

`building-agents` is the near-miss this skill is most often confused with (fixed DAG vs. the model picking its own next step), and it was missing from the old boundary.

## Body — what went

- **The "Core stance" bullet bank.** Three of its four rules were restated in full, with more detail, by the sections that own them: timeout + bounded retries in *Reliability mechanics*, the step-N → step-N+1 structured contract in *Design the chain as a typed DAG*, router-in-front in *Put a gateway/router in front*. Repeating a rule in a preamble and again at point of use is the "repeating yourself across layers" pattern the brief targets.
- The fourth bullet — *cache and tier before you tune prompts* / *the cheapest call is the one you never made* — was **not** restated anywhere, so it was preserved and moved to the head of the caching section, where it governs.
- The stance sentence itself survives, folded into the intro together with the outage / rate-limit / latency-tail rationale the deleted bullets carried.

## Body — what stayed, and why

- **The eight-row anti-patterns table.** Required by the rubric, and a different animal from a rationalization bank: every row names a concrete failure mode (naked call, `while True` retry, retrying a side-effecting step, weak-embedder semantic cache, schema-valid ≠ correct) with the bite and the fix. Already framed as `Anti-pattern | Why it bites | Do instead` — no STOP / excuse wrapper to strip.
- **The "Do you even need a pipeline?" routing table.** The flow genuinely branches, and this table is what keeps the skill out of `rag` / `building-agents` / `structured-extraction` / `cost-tracking` territory.
- The two-layer caching table (prefix vs. semantic is the skill's central distinction), both bad/good code pairs, the LiteLLM `v1.83.3-stable` and Anthropic Haiku/Sonnet/Opus price anchors, the observability field list, and the Verify block.

## Checks

- `bash scripts/eval-lint.sh` → `llm-pipeline PASS`.
- All 5 `route_to` targets exist under `skills/` (`prompt-engineering`, `structured-extraction`, `rag`, `building-agents`, `cost-tracking`); so do all 8 `recommends` and every `../<sibling>/SKILL.md` link.
- Both `references/` files still linked inline at point of use (`references/litellm-router.md`, `references/caching-layers.md`).
- `scripts/verify.sh` already executable; exits 0 on an empty target.

### DIFF
--- a/skills/llm-pipeline/SKILL.md
+++ b/skills/llm-pipeline/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: llm-pipeline
-description: "Use when wiring several LLM calls into one production flow, when random 429s/timeouts/provider outages take a feature down, when adding a fallback model or a gateway in front of OpenAI/Claude, or when the bill explodes from uncached or flagship-everywhere calls. Triggers: 'chain these LLM calls', 'step 1 extracts step 2 summarizes', 'add a fallback if GPT fails use Claude', 'route easy requests to Haiku escalate to Opus', 'put a LiteLLM gateway in front', 'our bill tripled because we re-call the model on every request', 'semantic cache the responses', 'encadenar llamadas y cachear respuestas', 'cadena de crides a l'LLM amb fallback'. NOT prompt wording (that is prompt-engineering)."
+description: "Use when wiring several LLM calls into a production flow: typed contracts between steps, a router/gateway so 429s, timeouts and outages fail over instead of taking you down, and cost control via caching, model tiers and abort caps. NOT single-prompt wording (that is `prompt-engineering`), NOT a model-driven tool loop (that is `building-agents`)."
 tags: [llm-orchestration, llm-gateway, fallbacks, prompt-caching, cost-control, litellm, reliability]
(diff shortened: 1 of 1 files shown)


ericrisco:

PR #175, opened 2026-07-29
refactor(prompt-engineering): Claude-5-era context pass

ericrisco:
Migration of `prompt-engineering` to the Claude-5-era rubric (`scripts/skill-rubric.md`, applied per `scripts/skill-migration-brief.md`).

## Numbers

| | before | after |
|---|---|---|
| description | 810 chars | 347 chars |
| SKILL.md | 10,234 bytes / 138 lines | 9,532 bytes / 128 lines |

## What went

- The nine-phrase trilingual `Triggers:` list. It is in context on every turn the skill is installed, invoked or not, and the body already relies on semantic matching.
- The top-of-file **"The one rule"** block. It pre-stated what the `Inline evals` section says a few screens later; the discipline now sits on the step it governs rather than as a banner above the prompt skeleton.
- The trailing **References** index. Both `references/` files are already pointed at inline at their point of use; the one detail the index carried that the inline mention did not (cases.yaml shape, assertion helpers, before/after diff runner) moved into the evals section, so no link was lost.

## What stayed, deliberately

- The **anti-patterns table**, all 7 rows. It was already framed `Anti-pattern | Why it bites | Do instead` — no STOP/excuse wrapper to strip — and it names concrete failure modes rather than restating rules.
- The **output-contract mechanism table**: the flow genuinely branches by provider (strict `json_schema` vs strict tool use vs JSON mode vs parse-it-yourself).
- The block skeleton with its per-block *why*, and the bad/good few-shot pair (judgment by demonstration).
- Every dated, load-bearing claim, verbatim: the OpenAI <0.1% schema-failure figure, the Anthropic prefill-returns-400 migration note, DSPy 3.2.1, GEPA/ICLR 2026. No recommendation, version, figure, or command was changed. This skill has no result envelope and none was invented.

## Also fixed in passing

- `agent-eval` and `agent-safety` were mentioned in bare backticks; they are now real `../<id>/SKILL.md` links (both resolve).
- The description now carries a second boundary against `agent-eval` — the old one advertised "writing a tiny eval set" with no line between this skill's inline cases and the standing golden-set/CI harness next door.
- Reworded the `evals/cases.yaml` `why` that quoted the deleted section title.

## Verification

`bash scripts/eval-lint.sh` → PASS for this skill. Both `references/` files linked from the body. `scripts/verify.sh` untouched, executable, exits 0 on an empty target. `manifest.json` deliberately not touched (derived; orchestrator regenerates). `npm test` / `npm run validate` not run — no `node_modules` in this worktree, per the brief.

### DIFF
--- a/skills/prompt-engineering/SKILL.md
+++ b/skills/prompt-engineering/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: prompt-engineering
-description: "Use when one prompt has to give the same right answer across reruns, models, and hostile inputs — a prompt that works sometimes, forcing valid JSON or a fixed schema out of a model, choosing how many few-shot examples and which ones, hardening a prompt against injected instructions, writing a tiny eval set before changing a prompt, or porting a prompt to a new model without behavior drift. Triggers: 'the prompt works sometimes', 'force valid JSON from the model', 'how many few-shot examples', 'outputs drift between reruns', 'the model ignores my instructions on long inputs', 'users can paste instructions into my input field', 'el prompt a veces falla', 'forzar JSON válido del modelo', 'el model no segueix les instruccions'. NOT the agent loop, tool wiring, or RAG retrieval (that is building-agents)."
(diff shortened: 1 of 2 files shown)


ericrisco:

PR #180, opened 2026-07-29
refactor(no-code-app): Claude-5-era rubric pass — description 537→346, body 154→146 lines

ericrisco:
Claude-5-era rubric pass on `no-code-app`. Context engineering only — no recommendation, figure, tier, cap or source was changed.

## Numbers

| | Before | After |
|---|---|---|
| `description` | 537 chars | **346 chars** |
| `SKILL.md` | 11847 bytes / 154 lines | **11325 bytes / 146 lines** |

## Description

Dropped the seven-phrase trilingual `Triggers:` list — it is in context on every turn the skill is installed, invoked or not, and the model matches by meaning without it.

Replaced the single vague `NOT a coded web app (that is nextjs)` with the two boundaries this skill is actually confused with:

> Use when building an app on a no-code/low-code platform (Bubble, Softr, Glide over Airtable or Xano) — picking the platform, modeling the data, keeping usage-metered cost predictable, and planning the exit before lock-in. NOT a hand-written coded app (that is `nextjs`), NOT wiring SaaS tools together with no app UI (that is `automation-flows`).

Both siblings verified present under `skills/`.

## What went

- **"The one rule" heading and its bullet bank.** The trade it stated (speed-to-first-user vs logic ceilings and scale economics) is now prose in the intro where it is read. The lock-in figures it carried were already repeated in "Plan the exit" — classic repeating-yourself-across-layers — so they now appear once, at the point where they drive a decision. Nothing lost: `~25-30% rewritten / $50K-$250K` moved into step 4, `~68% no code export` was already in step 1 and in the anti-patterns table.
- **The trailing `## References` index.** `references/platform-limits.md` is linked inline twice (platform table, cost-ceiling section), and the `stripe` pointer moved up next to the `payments` row in the data model. Every link survives at point of use; the reference invariant holds.

## Fixed while in here

`shopify` and `wordpress` sat in bare backticks in the Step 0 route table while every neighbouring row was a real link, and `react` was plain text. All three now link to skills that exist.

## What stayed, on purpose

- **The full 10-row anti-patterns table.** Concrete failure modes in a Bad/Good/Why framing — required by the rubric, and the only place the metering footguns are collected for skimming. No borrowed-urgency or excuse framing to strip.
- **All the decision tables** (route-out gate, platform-by-billing-model, backend-by-logic-complexity) and **both worked Bad → Good pairs**. The flow genuinely branches at each table; the pairs teach judgment by demonstration.
- **No result envelope was added** — this skill never had one.

## Verified

- `bash scripts/eval-lint.sh | grep '^no-code-app'` → PASS
- Frontmatter parses as YAML; `description` 346 chars; `origin: risco`; `name` matches the directory
- Every `../<sibling>/SKILL.md` link resolves (`nextjs`, `react`, `shopify`, `wordpress`, `automation-flows`, `spreadsheet-ops`, `notion-connector`, `stripe`)
- Every `references/` file is linked from the body
- `manifest.json` untouched; no `npm test` / `npm run validate` run in the worktree (no `node_modules`)
- `evals/cases.yaml` needed no change: no `why` quoted a deleted trigger phrase, and all five `route_to` targets name real skills

### DIFF
--- a/skills/no-code-app/SKILL.md
+++ b/skills/no-code-app/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: no-code-app
-description: "Use when building an app on a no-code/low-code platform (Bubble, Softr, Glide, with Airtable/Xano backends) — picking the tool, modeling the data, taming usage/cost metering, or planning the exit before lock-in. Triggers: 'build a CRM without code', 'Bubble or Glide for my MVP', 'monta una app sin codi per reserves', 'my Bubble bill suddenly exploded' (workload-unit metering), 'Glide updates overage', 'how do I avoid getting locked into a no-code tool', 'when should I stop and write real code'. NOT a coded web app (that is nextjs)."
+description: "Use when building an app on a no-code/low-code platform (Bubble, Softr, Glide over Airtable or Xano) — picking the platform, modeling the data, keeping usage-metered cost predictable, and planning the exit before lock-in. NOT a hand-written coded app (that is `nextjs`), NOT wiring SaaS tools together with no app UI (that is `automation-flows`)."
 tags: [no-code, low-code, bubble, glide, softr, app-builder, mvp]
 recommends: [nextjs, shopify, wordpress, automation-flows, spreadsheet-ops, notion-connector, stripe]
 origin: risco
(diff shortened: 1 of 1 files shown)


ericrisco:

PR #181, opened 2026-07-29
refactor(invoicing): Claude-5-era context pass

ericrisco:
Context-engineering pass on `skills/invoicing/` per `scripts/skill-migration-brief.md`. Format only — no recommendation, version, figure, or sourced claim was changed.

## Numbers

| | Before | After |
| --- | --- | --- |
| `description` chars | 946 | 343 |
| body bytes | 12175 | 10926 |
| body lines | 154 | 145 |

## Description

Dropped the 11-phrase trilingual `Triggers:` list. That list sits in context on every turn the skill is installed, invoked or not, and the model already matches by meaning. What replaced it keeps the job statement and sharpens the boundary to the three siblings this is actually confused with — `bookkeeping` (recording it), `stripe` (SDK plumbing), `pricing` (setting the amount) — in backticks, per the current convention.

Before/after discrimination test: the new text still lets a reader pick this over `bookkeeping` and `stripe` on its own.

## What went

- `## Hand-off`, the trailing index. All seven destinations already appear in the "Route out before you start" table at point of use; `bookkeeping` additionally appears in the lifecycle table and the anti-patterns table. Nothing was dropped — `finance-ops` only ever lived in the table.
- The "You are not the accountant... not the engineer" sentence — the same boundary the description and the route table already state.
- The duplicated half of the numbering warning under the document-type table. The full rule (never delete, gap = audit flag, void vs credit note) stays where the decision is made, in the lifecycle section.

## What stayed, on purpose

- The anti-patterns table, all eight rows and three columns untouched. It names concrete failure modes, not rules restated from above.
- All four branching tables: document type, invoice states, route-out, legally-valid checklist.
- Every sourced claim verbatim: EU VAT invoicing fields, the per-country e-invoicing mandate dates (BE/FR/DE/ES/IT/PL, ViDA 2030), Directive 2011/7/EU (ECB reference + 8pp, EUR 40 fixed recovery cost), stripe-node v19.1.0 and its status enum, and the code sample.
- Both `references/` links, so neither file becomes dead weight.
- No result-envelope block was invented — this skill never had one.

## Verification

- `bash scripts/eval-lint.sh` → `invoicing PASS (should_trigger=7, should_not_trigger=6, capability=1)`
- Frontmatter parses as YAML; `description` 343 chars, single-line, well under the 1024 hard limit.
- Every `route_to` in `evals/cases.yaml` names a real skill (bookkeeping, webhooks, pricing, contracts, proposals, e-signature) — checked against `skills/`.
- `scripts/verify.sh` already executable (`755`).

### DIFF
--- a/skills/invoicing/SKILL.md
+++ b/skills/invoicing/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: invoicing
-description: "Use when someone needs to cut a legally-valid bill or quote, move it through its get-paid lifecycle, or chase money that hasn't arrived — billing completed work, drafting a quote before a job, an invoice weeks overdue, what must legally be on the page, or readiness for a structured e-invoicing mandate. Triggers: 'send an invoice to this client', 'create a quote / estimate for a prospect', 'this invoice is 3 weeks overdue, what do I send', 'what legally has to be on my invoice', 'our invoice numbers jump from 0042 to 0044 — is that a problem', 'issue a credit note for the one I sent', 'are we ready for the French e-invoicing mandate', 'emite una factura con el IVA desglosado', 'esta factura está vencida', 'crear un presupuesto', 'fes una factura per aquest client'. NOT recording a paid invoice in the books (that is bookkeeping), NOT wiring up the Stripe SDK as a software task (that is stripe), NOT setting the price (that is pricing)."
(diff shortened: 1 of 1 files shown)


ericrisco:

PR #183, opened 2026-07-29
refactor(roast-me): trigger-phrase list out of the description, intro de-duplicated

ericrisco:
Context-engineering pass on `roast-me` per `scripts/skill-migration-brief.md` and `scripts/skill-rubric.md`.

## Numbers

| | Before | After |
|---|---|---|
| `description` chars | 612 | **347** (target ≤350, hard limit 1024) |
| `SKILL.md` bytes | 5873 | **5387** |
| `SKILL.md` lines | 124 | **122** |
| `scripts/eval-lint.sh` | PASS | **PASS** (should_trigger=7, should_not_trigger=5, capability=2) |

## What went

- **The `Triggers:` list.** Nine phrases across English, Spanish and Catalan, paid for on every turn the skill is installed whether or not it fires. The model matches by meaning; the skill's own eval cases include Spanish and Catalan prompts precisely to prove routing does not need the keywords spelled out.
- **The intro sentence that re-listed the phase outputs** (dual scores, habit list with named techniques, trend over time) two lines before the five phases specify each of them. Repeating yourself across layers is the first pattern the Jul 2026 context-engineering rules name as not paying.
- **A duplicated instruction in Phase 5** — "If history exists, append the trend line" appeared again four lines after the block that already said it — plus the stray trailing `Done.`

## What stayed, deliberately

- **The whole five-phase flow** and every instruction inside it: extractor invocation and `$ARGUMENTS` parsing, the ~30-prompt batching, the impact filter, the aggregation keys, the roast subagent payload, the history JSON shape and the trend-line format. This was a format pass, not a content rewrite — no recommendation, figure, path or command changed.
- **The stance sentence** ("your target is not the user's code — it is their *prompting behaviour*"). It reads like a restatement of the boundary, but it is what keeps a loaded run from drifting into a code review, so it earns its line.
- **The `effective_error_rate` rule** — the skill's single most distinctive piece of judgment.
- **The shared `Orientación` block**, byte-identical to the one in every other migrated skill (`harness`, `implement`, `verify`, `ship`, …). Uniformity beats a local trim.

Boundary reworded into the catalog's `NOT <x> (that is `sibling`)` shape. Both siblings verified to exist: `skills/code-review/`, `skills/prompt-engineering/`.

> Use when someone wants an honest, comedic audit of their own prompting — reads their past agent transcripts, scores prompt quality and compute efficiency, names their worst habits, builds a model-selection cheat sheet. NOT reviewing code an agent produced (that is `code-review`), NOT prompting technique in general (that is `prompt-engineering`).

## evals

Three `why` fields justified their case by pointing at the trigger list that no longer exists ("the skill's own trigger phrase", "the description lists this phrasing explicitly"). Reworded to justify by meaning instead. Cases, counts and every `route_to` are untouched — and all five targets were verified real: `code-review`, `prompt-engineering`, `analyze`, `chatbot`, `llm-pipeline`. The three `recommends` ids are real too (`prompt-engineering`, `code-review`, `context-budget`).

## Notes for the orchestrator

- `roast-me` has **no `references/` directory** — its progressive-disclosure payload is `prompts/analyze.md`, `prompts/compute.md`, `prompts/roast.md`, and all three are still linked inline from the body (Phases 2, 3, 4). No `scripts/verify.sh`, so gate 7 does not apply.
- **Content gap left alone on purpose:** the body has no anti-patterns table, which rubric dimension 3 asks for. Adding one would mean inventing content in a format pass, so it is flagged rather than fixed.
- `manifest.json` untouched, per the brief.

### DIFF
--- a/skills/roast-me/SKILL.md
+++ b/skills/roast-me/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: roast-me
-description: "Use when you want honest, comedic feedback on how you prompt — analyzing your own past sessions to score prompt quality and compute efficiency, surface your worst habits, and generate a model-selection cheat sheet. Triggers: 'roast me', 'roast my prompting', 'audit my prompting habits', 'how good are my prompts', 'am I prompting well', 'what are my bad prompt habits', 'analiza mis prompts', 'puntua els meus prompts', 'how much am I wasting on model costs'. NOT reviewing your code or output quality (use code-review for that) and NOT a general prompt-engineering tutorial (use prompt-engineering for that)."
+description: "Use when someone wants an honest, comedic audit of their own prompting — reads their past agent transcripts, scores prompt quality and compute efficiency, names their worst habits, builds a model-selection cheat sheet. NOT reviewing code an agent produced (that is `code-review`), NOT prompting technique in general (that is `prompt-engineering`)."
 tags: [prompting, self-audit, compute-efficiency, ai-hygiene, learning]
 recommends: [prompt-engineering, code-review, context-budget]
 profiles: [full]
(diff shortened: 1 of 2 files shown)


ericrisco:

PR #185, opened 2026-07-29
refactor(power-automate): Claude-5-era rubric pass

ericrisco:
Context-engineering pass on `skills/power-automate/` per `scripts/skill-migration-brief.md`. No content rewrite: no recommendation, version, endpoint, command or figure changed.

| | before | after |
|---|---|---|
| description | 997 chars | **350 chars** |
| SKILL.md | 14141 bytes / 169 lines | **12630 bytes / 160 lines** |

**eval-lint: PASS** — `power-automate  PASS (should_trigger=6, should_not_trigger=5, capability=1)`

## Description

Dropped the five-phrase `Triggers:` list — it is in context on every turn the skill is installed, invoked or not, and the body already routes by meaning. Dropped the five-way `NOT` chain down to the two boundaries that actually discriminate: designing the flow definition (`automation-flows`) and platform choice by billing model (`automation-strategy`). The other three (`api-connector-builder`, `webhooks`, `google-workspace`) are routing detail and now live in the body, read at the point of use.

## What went

- **`## References` tail index.** All three `references/` files were already linked inline — `entra-auth-setup.md` at the auth paragraph, `flowstudio-mcp-and-limits.md` in the API-vs-MCP table and the debug step, `dataverse-web-api.md` at the `clientdata` trap. What the index added over the bare link (endpoint cheat-sheet, OData headers, error handling) is folded into the inline mention, so the invariant holds and nothing about the package is lost.
- **`## Related skills`.** A third statement of the same boundaries. Merged with `## Microsoft 365 ecosystem fit` into one `## Boundaries` section that keeps every claim and every sibling link, written as three concrete "not this" cases instead of a link list. All five `../<sibling>/SKILL.md` links resolve.

## What stayed, deliberately

- The three honesty facts up front (My Flows are not code-manageable, `api.flow.microsoft.com` is unsupported, the Power Platform API is moving fast). Heaviest-weighted dimension; load-bearing, not padding.
- The API-vs-MCP decision table and the `workflow` column table — the flow genuinely branches on both.
- The full lifecycle with its `curl` calls, the ExportSolution-before-`DELETE` irreversibility rule, and the pre-flight checklist (kept in line with the other migrated automation skills, e.g. `automation-strategy`).
- The gotcha table, retitled `## Anti-patterns` with `Anti-pattern | Why it bites | Do instead` to match the catalog convention. **All seven rows preserved verbatim** — they name concrete failure modes, not rules restated from above.

## Notes

- `evals/cases.yaml` needed no change: every `route_to` names a real sibling, no `why` quotes a deleted trigger phrase, and the "no first-party MS-Graph skill" claim is still true (no `microsoft-365`/`ms-graph` skill in the catalog).
- No `scripts/` dir in this skill, so no `verify.sh` gate applies. `manifest.json` untouched.

### DIFF
--- a/skills/power-automate/SKILL.md
+++ b/skills/power-automate/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: power-automate
-description: "Use when operating Microsoft Power Automate cloud flows programmatically — creating, enabling, updating, listing or deleting flows through the Dataverse Web API (the workflow table, category 5) with Entra ID / OAuth2 auth, or reading action-level run history via the third-party FlowStudio MCP. Triggers: 'create a Power Automate cloud flow with the Dataverse API', 'enable my flow by code / set statecode to 1', 'why can't I create or edit a My Flow programmatically', 'PATCH the workflow clientdata', 'automatizar el despliegue de un flujo de Power Automate desde código'. NOT designing the flow or authoring its definition (automation-flows), NOT choosing a platform by billing model (automation-strategy), NOT a generic typed REST client with auth/pagination/backoff (api-connector-builder), NOT the endpoint receiving an HTTP-trigger POST in your app (webhooks), NOT scripting Microsoft 365 / Graph mail & calendar directly (no first-party skill — api-connector-builder, or google-workspace)."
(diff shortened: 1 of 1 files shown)


ericrisco:

PR #186, opened 2026-07-29
refactor(linkedin-outreach): Claude-5-era description and body pass

ericrisco:
Context-engineering pass on `linkedin-outreach` per `scripts/skill-migration-brief.md`. No substance changed — no recommendation, figure, price, path, or command was touched.

## Numbers

| | Before | After |
| --- | --- | --- |
| `description` chars | 841 | 356 |
| SKILL.md bytes | 12,759 | 11,321 |
| SKILL.md lines | 174 | 154 |

## Description

Was a seven-phrase `Triggers:` list across English, Catalan and Spanish, followed by a restatement of what the skill owns. Now the discriminating shape — what it does, then the two boundaries that actually get confused:

> Use when running LinkedIn social selling as a repeatable motion — Sales Navigator saved searches and trigger alerts, weekly request budgets, fixing a low acceptance rate, turning profile views and post engagers into warm threads, logging every touch. NOT the note or DM copy (that is `cold-outreach`), NOT the deal after a reply (that is `sales-pipeline`).

Both named siblings exist under `skills/`.

## What went

- **"When to use"** — eight example prompts that restated the description more slowly.
- **The doubled boundary section.** "When NOT — the boundary that gets fought over" plus a second bullet list of other routes became one `## Boundary` section: the message-vs-system split against `cold-outreach` (the genuinely contested line, kept) and the four secondary routes on one line. All five sibling links survive, and the three that were bare backticks in the intro are now real `../<id>/SKILL.md` links.
- **A repeated statistic.** Loop phase 2 quoted the ~20–30% → ~60%+ acceptance lift, which is stated again where the ritual is actually prescribed. Kept at the point of use only.
- **"Explicit stop"** — folded into the handoff step it governs, per the brief's rule that a rule in context is followed and a rule in an appendix is skimmed.
- **The health-rules paragraph** that restated the weekly-budget table's own "Health rule" column; the one fact it added (withdraw stale pending requests) was preserved.

## What stayed, on purpose

- **The Anti-patterns table, all six rows**, in its existing `Bad | Why it bites | Do instead` columns. It names concrete failure modes rather than shouting rules already stated above it, so it is the table the rubric requires — not a rationalization bank.
- **All four decision tables**, because the flow genuinely branches on each: Free/Core/Advanced buy math with 2025 pricing, the note-or-no-note paradox with its measured accept-vs-reply rates, the weekly request budget by account type, and the signal → play mapping.
- The worked bad/good pitch-on-accept pair, the five-phase loop with its per-phase *why*, the ledger schema and weekly-review formulas, and the daily 30-minute checklist.
- Both `references/` files remain linked inline (`sales-navigator-playbook.md`, `ledger-schema.md`).
- No result-envelope block existed and none was invented.

## Verification

- `bash scripts/eval-lint.sh | grep '^linkedin-outreach'` → **PASS** (should_trigger=6, should_not_trigger=5, capability=1).
- Frontmatter parses as YAML; `name` matches the directory, `origin: risco` present; description 356 chars, well under the 1024 limit.
- All five `../<sibling>/SKILL.md` links resolve; both `references/` files are linked from the body.
- `scripts/verify.sh` is executable and exits 0 with an advisory warning on an absent ledger.
- `evals/cases.yaml` untouched and needed no rewording — every `route_to` names a real skill and no `why` quoted a deleted trigger phrase.
- `manifest.json` deliberately not touched; per the brief the orchestrator regenerates it.

### DIFF
--- a/skills/linkedin-outreach/SKILL.md
+++ b/skills/linkedin-outreach/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: linkedin-outreach
-description: "Use when running LinkedIn-native social selling as a repeatable motion: building Sales Navigator saved searches and trigger alerts, fixing a low connection-acceptance rate, designing a daily social-selling routine, turning profile views / post engagers / accepted connections into warm conversations, and logging every touch so the loop is measurable. Triggers: 'set up a Sales Nav saved search and alert me on job changes', 'how many connection requests per week without getting flagged', 'my acceptance rate is in the toilet', 'someone viewed my profile, what is the play', 'turn the people who liked my post into conversations', 'munta'm una rutina de social selling a LinkedIn', '¿cuántas invitaciones puedo enviar por semana'. This owns the loop and its ledger, NOT writing the cold connection note or DM copy (that is cold-outreach)."
+description: "Use when running LinkedIn social selling as a repeatable motion — Sales Navigator saved searches and trigger alerts, weekly request budgets, fixing a low acceptance rate, turning profile views and post engagers into warm threads, logging every touch. NOT the note or DM copy (that is `cold-outreach`), NOT the deal after a reply (that is `sales-pipeline`)."
 tags: [linkedin, social-selling, sales-navigator, outreach, pipeline]
 recommends: [cold-outreach, lead-gen, sales-pipeline, linkedin-strategy, linkedin-content]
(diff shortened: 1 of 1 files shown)


ericrisco:

PR #188, opened 2026-07-29
refactor(react-native): Claude-5-era description + body pass

ericrisco:
## Numbers

| | Before | After |
|---|---|---|
| `description` | 722 chars | **348 chars** |
| Body | 12264 bytes / 220 lines | **11715 bytes / 210 lines** |
| `scripts/verify.sh` | exit 1 (stale check) | exit 0 on the skill and on an empty target |

## What went

- **The `Triggers:` phrase list** — 9 quoted phrases across English, Catalan and Spanish. The description is in context on every turn the skill is installed, invoked or not; the model matches by meaning, so the list was pure per-turn cost. This is the bulk of the 722 → 348 saving.
- **The trailing `## Cross-references` index.** `../expo/SKILL.md` was already stated twice inline (verb table + anti-patterns row), and the flutter boundary is in the description. The two pointers that existed *only* in the index moved to their point of use: `../design/SKILL.md` next to safe-area/layout, `../debug/SKILL.md` at the end of the debugging list. No link was dropped without a replacement.
- **The `## What this skill owns` heading** and the sentence ("the pipeline around it … is not yours") that restated what the verb table's second row already says.

## What stayed, on purpose

- **The verb table.** `react-native` vs `expo` is the boundary users actually get wrong, and it is a genuine branch — not a restatement of the description.
- **The anti-patterns table**, all ten rows, already in `Anti-pattern | Why it hurts | Do instead` shape with no borrowed urgency/rationalization framing.
- **Every load-bearing claim verbatim**: the New Architecture mandate (RN 0.82 / SDK 55), the SDK→RN map, the FlashList ~100-item threshold, Reanimated 3 vs 4, React Compiler GA in SDK 54, the router and native-module decision tables, and all bad/good code pairs. This was a context-engineering pass, not a content rewrite.

## One in-scope script fix

`scripts/verify.sh` check 4 required the description to **carry** a `Triggers:` label — it encoded the exact convention this migration removes, so it failed by design the moment the skill was migrated. It now checks the `Use when/to` lead and the `NOT … (that is <sibling>)` boundary, and flags a `Triggers:` list as the violation it now is. `flag()` also returns 0 so a message-only finding can no longer trip `set -e` and silently skip the remaining checks.

## Verified

- Both `references/` files are still linked inline from the body (`native-modules.md`, `performance-debugging.md`).
- Every `../<sibling>/SKILL.md` link resolves (`expo`, `design`, `debug`).
- Every `route_to` in `evals/cases.yaml` names a real skill (`expo`, `flutter`, `nextjs`, `ship`) — no stale "not in this catalog" claims, nothing left quoting a deleted trigger phrase.
- `manifest.json` untouched.

### DIFF
--- a/skills/react-native/SKILL.md
+++ b/skills/react-native/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: react-native
-description: "Use when writing the JS/TS code inside a React Native or Expo app — screens, navigation, lists, animation, platform-conditional UI, data/offline state, and native-module authoring — and when killing runtime jank or render bugs. Triggers: 'scaffold an Expo Router app', 'FlashList vs FlatList', 'auth-gated route groups', 'gesture reattaches every render', 're-render storm on every keystroke', 'deep link lands on the wrong screen', 'write an Expo Module wrapping a native SDK', 'la llista pateix lag quan faig scroll', 'mi app va fluida en iOS pero a tirones en Android'. NOT eas build/submit/OTA update/prebuild/config-plugin (that is expo), NOT a Dart app (that is flutter), NOT web React/DOM (that is react or nextjs)."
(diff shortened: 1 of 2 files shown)


ericrisco:

PR #189, opened 2026-07-29
refactor(meeting-notes): cut the trigger-phrase list and the cross-layer repeats

ericrisco:
Context-engineering pass on `meeting-notes` against the current rubric. Format and layering only — no recommendation, figure, version, or citation changed.

## Numbers

| | Before | After |
|---|---|---|
| `description` chars | 777 | **360** |
| SKILL.md bytes | 13610 | **12497** |
| SKILL.md lines | 168 | **153** |

## Description

The old one ended in a four-language `Triggers:` keyword list — paid for on every turn the skill is installed, invoked or not, while the body already routes by meaning. Deleted. What survives is what discriminates: what it does, when it fires, and the boundary against its nearest sibling.

> Use when a meeting just ended and a transcript or rough notes must become a record the absent can act on — decisions with their why, actions with an owner and a real date, open questions, a recap inside 24–48h — or to clean an AI-notetaker draft that invented tasks or swapped speakers. NOT the durable ADR with alternatives weighed (that is decision-records).

## What went

- **`What this owns vs. what it does not` → a `Not this skill` routing table.** All seven delegations (`decision-records`, `sop-builder`, `project-ops`, `calendar-scheduling`, `notion-connector`, `automation-flows`, `document-processing`) survive with their distinguishing clause. The trailing "Boundary in one line" restated the first row *and* the description's NOT clause — gone.
- **The three action-item bullets** folded into the `verb, owner, by <real date>` line that introduced them. Same content ("Draft…", not "the team", never "ASAP"); the restate → demonstrate → enumerate layering is what went, since the Bad → Good table directly below already teaches each rule.
- **The "one sentence of context" rule** moved out of the skeleton section into the Decisions step it actually governs, carrying its citation with it.
- **The consent decision checklist** was a line-for-line restatement of the four consent bullets above it. Dropped only after verifying each item still lands somewhere: the state list / GDPR / privilege-waiver / redaction details are in the bullets, and "hand-note sensitive meetings" is an anti-patterns row.
- **The trailing References index** folded into an inline pointer in the skeleton section, so `references/templates.md` is still linked (twice) and the four templates are still named where a reader would want them.

## What stayed, deliberately

- **The ten-row anti-patterns table**, untouched — it was already `Anti-pattern | Why it fails | Do instead` with no borrowed urgency framing, and it names concrete failure modes rather than restating rules.
- **Both Bad → Good tables** (action items, decisions) — worked pairs that teach judgment by demonstration.
- **The source-branch decision table** (live meeting / AI transcript / raw bullets) — the flow genuinely branches there.
- **All grounding verbatim**: Whisper WER figures with the arXiv citation, the Justia 50-state survey with its "the canonical list is contested" hedge, the ABA GPSolo reference, and every `accessed 2026-06-02` date.
- **No result envelope was added** — this skill never had one.

## Also

- Linked the one sibling still sitting in bare backticks (`decision-records` in the ADR hand-off line).
- `evals/cases.yaml` untouched: all four `route_to` targets are real skills, no stale "not in this catalog" claims, and no `why` quoted a deleted trigger phrase.
- `manifest.json` untouched, per the migration brief.

### DIFF
--- a/skills/meeting-notes/SKILL.md
+++ b/skills/meeting-notes/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: meeting-notes
-description: "Use when a meeting, standup, or call just ended and you have a transcript or rough notes to turn into a durable record — extract the decisions with their why, assign action items with an owner and a real due date, flag open questions, and ship a one-screen recap within 24–48h. Also when an AI-notetaker transcript hallucinated tasks or mixed up speakers and you must clean it before sending. Triggers: 'pull the action items out of this transcript and tell me who owns what by when', 'the Otter transcript mixed up who said what, clean it up so I can send it', 'resume la reunión de hoy y saca los próximos pasos con responsable y fecha límite', 'fes una acta de la reunió'. NOT the durable decision record with full alternatives and review cadence (that is decision-records)."
+description: "Use when a meeting just ended and a transcript or rough notes must become a record the absent can act on — decisions with their why, actions with an owner and a real date, open questions, a recap inside 24–48h — or to clean an AI-notetaker draft that invented tasks or swapped speakers. NOT the durable ADR with alternatives weighed (that is decision-records)."
 tags: [meeting-notes, action-items, decisions, recap, transcript-cleanup, minutes, standup, business-ops]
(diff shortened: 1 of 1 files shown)


ericrisco:

PR #192, opened 2026-07-29
refactor(github-actions): Claude-5-era description and body pass

ericrisco:
Context-engineering pass on `github-actions` against `scripts/skill-rubric.md`. Format only — no recommendation, version, command or figure changed.

| | Before | After |
| --- | --- | --- |
| `description` | 865 chars | **348 chars** |
| `SKILL.md` | 256 lines / 14 550 B | **244 lines / 13 375 B** |

## What went

- The `Triggers:` list — eight quoted phrases in English and Catalan. It sat in every installed user's context on every turn, and the model matches by meaning anyway. The description now carries only what it does, when to use it, and the three `NOT … (that is <sibling>)` boundaries.
- `## Use this when` (7 bullets) — a restatement of the description one screen below it.
- `## Not this when` as prose — folded into a single routing table. Every destination survives (docker, git-workflow, ship, deployment, vercel/aws-essentials, secure-coding) and each is now a resolving `../<id>/SKILL.md` link; previously `docker` and `git-workflow` were named in bare text only.
- The third copy of "keep `GITHUB_TOKEN` read-only, widen per job". It now appears once, next to the `permissions:` line it governs, with the `packages: write` / `id-token: write` examples folded in — and once more as an anti-pattern row, which is a failure mode, not a restatement.
- The standalone `npm ci` bullet; its reasoning moved onto the comment on the step itself.

## What stayed, deliberately

- The **anti-patterns table**, all nine rows. Concrete failure modes (tag-pinned third-party action, `repo:ORG/*` OIDC trust, `pull_request_target` + PR head, keyless cache key…), not rules already stated above — the rubric requires it and it earns its lines.
- The trigger-selection table and the reusable-workflow vs composite-action table: the flow genuinely branches there.
- The bad/good pairs for cache keys, OIDC credentials and SHA pinning — they teach the judgment by demonstration.
- Every dated, load-bearing fact: `checkout` v6.0.2 / `setup-node` v6.4.0, Node 20 runner removal September 2026, Node 24 default since June 2026, the tj-actions/changed-files 2025 compromise.

## Verification

- `bash scripts/eval-lint.sh` → `github-actions PASS (should_trigger=7, should_not_trigger=5, capability=1)`.
- Frontmatter parses as YAML; `description` 348 chars; `name` matches the directory; `origin: risco`.
- All seven `../<sibling>/SKILL.md` links resolve; both `references/` files (`caching-and-matrix.md`, `oidc-deploys.md`) are still linked inline at point of use.
- `scripts/verify.sh` already executable, exits 0 on an empty target.

### DIFF
--- a/skills/github-actions/SKILL.md
+++ b/skills/github-actions/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: github-actions
-description: "Use when authoring or fixing GitHub Actions CI/CD — workflows under .github/workflows, jobs, the runner matrix, dependency caching, secrets and OIDC cloud deploys, environments and approval gates, reusable and composite workflows. Triggers: 'write a CI workflow', 'my GitHub Actions builds are slow / add caching', 'run the tests on a matrix of node versions', 'deploy to AWS without storing access keys' (OIDC), 'every push spawns a run and the old ones keep finishing' (concurrency), 'pin my actions / supply-chain hardening after the tj-actions thing', 'munta el desplegament continu amb GitHub Actions cap a producció', 'per què el GITHUB_TOKEN té permisos d'escriptura per defecte'. NOT writing the Dockerfile or image build strategy (that is docker), NOT the branching/merge model (that is git-workflow), NOT release readiness or the changelog (that is ship)."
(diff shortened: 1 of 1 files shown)


ericrisco:

PR #193, opened 2026-07-29
refactor(nodejs): Claude-5 context-engineering pass

ericrisco:
Applies `scripts/skill-rubric.md` to `nodejs`. Context-engineering pass only — no recommendation, version, command or figure changed.

## Numbers

| | Before | After |
| --- | --- | --- |
| `description` | 777 chars | **341 chars** (-56%) |
| `SKILL.md` | 13732 bytes / 242 lines | **12832 bytes / 230 lines** |

The description number is the one that matters: it is in context on every turn the skill is installed, invoked or not.

## Description

Before carried an eight-phrase bilingual `Triggers:` list. After is discriminative — what it does, plus three `NOT … (that is <sibling>)` boundaries naming real siblings:

> Use when building or operating a plain Node.js / Express 5 backend service: project layout, async correctness, central error middleware, fail-fast config, graceful shutdown on SIGTERM. NOT DI modules/providers/guards (that is `nestjs`), NOT the type system or tsconfig (that is `typescript`), NOT REST contract design (that is `api-design`).

## What went

- **`## Use this when`** (6 bullets) — restated the description line for line.
- **`## References & siblings`** tail index — both reference files were already pointed to at their point of use, so the index was a second copy of links the reader had already passed.

## What was merged

`## Not this when` and the tail's sibling line were two overlapping route tables saying the same thing in two places. They are now one `## Boundary` section carrying **every row from both** — including the `error-handling` routing that existed only in the tail, and the middleware-order-diagram note that existed only in the reference index (now folded into the inline mention at the point where mis-ordered middleware is being explained).

## What stayed, deliberately

- **The `## Decide first` table.** The flow genuinely branches — nestjs vs Express 5 vs `node:http` vs not-a-server — and the brief keeps decision tables where that is true.
- **The anti-patterns table, all nine rows.** Required by dimension 7, and a different animal from a rule bank: each row names a concrete failure mode with its consequence, not a rule restated from the section above.
- **The async rules, the Express 5 arity/order rule, and every bad/good pair.** These are the substance of the skill, and they already sit next to the step they govern rather than in an appendix.
- **The `scripts/verify.sh` note**, moved under its own `## Verify` heading.

No result-envelope block existed here and none was invented.

## Housekeeping

Bare-backtick sibling mentions — `api-design`, `error-handling`, `prisma-orm`, `drizzle-orm`, `docker`, `deployment`, `observability` — are now real `../<id>/SKILL.md` links, each verified to exist under `skills/`. Both `references/` files are linked inline, so neither is dead weight. `evals/cases.yaml` needed no change: its `why` fields describe the routing rather than quoting deleted trigger phrases, and no `route_to` makes a stale claim about the catalog. `scripts/verify.sh` was already executable. `manifest.json` deliberately untouched — it is derived and regenerated centrally.

### DIFF
--- a/skills/nodejs/SKILL.md
+++ b/skills/nodejs/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: nodejs
-description: "Use when building or structuring a plain Node.js / Express backend service — project layout, async correctness, central error handling, config validation, and graceful shutdown — the runtime and HTTP service, not a DI framework. Triggers: 'set up an Express 5 service', 'my async route error returns 200 with an empty body', 'where do controllers vs services go', 'centralize error handling middleware', 'shut down cleanly on SIGTERM without dropping requests', 'why does one unhandled promise rejection crash my server', 'monta un backend con Express y manejo de errores', 'apaga el servidor sin perder peticiones en vuelo'. NOT NestJS DI/modules/guards (that is nestjs); NOT the TS type system/tsconfig (that is typescript); NOT REST contract/versioning (that is api-design)."
+description: "Use when building or operating a plain Node.js / Express 5 backend service: project layout, async correctness, central error middleware, fail-fast config, graceful shutdown on SIGTERM. NOT DI modules/providers/guards (that is `nestjs`), NOT the type system or tsconfig (that is `typescript`), NOT REST contract design (that is `api-design`)."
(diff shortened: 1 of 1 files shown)


ericrisco:

PR #194, opened 2026-07-29
refactor(reporting): Claude-5-era description + body pass

ericrisco:
Migrates `reporting` to the `scripts/skill-rubric.md` rubric. Context-engineering pass only — no recommendation, version, figure or command changed.

## Numbers

| | Before | After |
|---|---|---|
| `description` | 1014 chars | **350 chars** |
| `SKILL.md` | 11695 bytes / 178 lines | **10841 bytes / 176 lines** |
| eval-lint | — | **PASS** (should_trigger=6, should_not_trigger=6, capability=1) |

## Description

Was a Symptoms paragraph plus a six-phrase trilingual `Triggers:` list — paid for on every turn the skill is installed, invoked or not, to do a keyword job the model already does by meaning. Now: what it does, when it fires, and three boundaries against siblings that exist under `skills/` (`dashboard`, `kpi-framework`, `analytics`).

> Use when a recurring report must ship itself on a cadence — weekly digest, monthly exec pack, board pack — via a fetch→narrate→render→deliver pipeline with a schedule and a freshness gate. NOT a live view people slice (that is `dashboard`), NOT picking which KPIs to track (that is `kpi-framework`), NOT a one-off investigation (that is `analytics`).

The trilingual eval prompts (`automatiza un informe mensual para dirección`, `munta un informe setmanal automàtic per al comitè`) are kept — they now test the semantic matching the skill relies on instead of a literal string it carried in context.

## Removed from the body

- The trailing `References:` index. `references/pipeline.md` is already linked inline at both points of use, so the tail entry was a duplicate link; the invariant that every `references/` file stays reachable from the body holds.
- Two sentences that restated anti-patterns rows word for word: "Do not email a dashboard link and call it a report" and "Never ship a raw metric dump with no narrative". The non-duplicated halves next to them survive.
- One clause of padding in the push-vs-dashboard rationale, and prose in `## Verify` — every check `verify.sh` actually performs is still named.

## Kept deliberately

- **The nine-row anti-patterns table, whole.** Required by dimension 3, and it is the real article rather than the rationalization variant: concrete failure modes (unpinned WeasyPrint reflowing a board pack, a template forked per recipient, an unread report left on a schedule), already columned `Anti-pattern | Why it bites | Do instead`, no STOP/excuse framing to strip.
- The four branching tables — report contract, push-vs-pull routing, artifact format, scheduler choice.
- The bad/good narrative pair, which teaches the judgment by demonstration.

## Also fixed

`dashboard`, `kpi-framework` and `data-cleaning` were named as routes in bare prose. Now linked, so the handoff the text promises can be taken. All nine `../<sibling>/SKILL.md` links resolve.

`manifest.json` untouched (derived). `npm test` / `npm run validate` not run — no `node_modules` in the worktree, per the brief.

### DIFF
--- a/skills/reporting/SKILL.md
+++ b/skills/reporting/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: reporting
-description: "Use when you need a recurring report that ships to stakeholders on a cadence — a weekly team digest, monthly exec pack, quarterly board summary, or recurring client performance report — and it must go out reliably without a human in the loop. Symptoms: a standing metrics email keeps breaking or shows stale numbers, the Monday digest is hand-assembled every week, leadership asks for 'the same report but every month', or a client expects a PDF on a schedule. Triggers: 'set up a weekly automated report emailed to the exec team', 'build a monthly PDF report for our biggest client on a schedule', 'send the board a quarterly summary every quarter', 'our Monday metrics email keeps going out with last week's numbers — make it reliable', 'automatiza un informe mensual para dirección', 'munta un informe setmanal automàtic per al comitè'. NOT a live screen people open to slice metrics (that is dashboard), NOT deciding which KPIs to track (that is kpi-framework), NOT a one-off investigation (that is analytics)."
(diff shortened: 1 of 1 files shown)


ericrisco:

PR #195, opened 2026-07-29
refactor(nestjs): Claude-5-era description and body pass

ericrisco:
Context-engineering pass on `skills/nestjs/` against `scripts/skill-rubric.md`. No content, version, command or recommendation changed.

**Description:** 671 → 355 chars. Dropped the eight-phrase bilingual `Triggers:` list (the model matches by meaning; that list was paid for on every turn the skill is installed) and added an explicit two-way boundary: NOT a bare Express/Fastify service with no DI (`nodejs`), NOT framework-agnostic REST design (`api-design`). Both siblings verified present under `skills/`.

**Body:** 12618 → 11787 bytes (226 → 218 lines).

Removed:
- the `Triggers:` phrase list (frontmatter)
- the `## Use this when` bullet list — a third restatement of the routing signal; each bullet is already covered at point of use (provider forms in "Providers & DI", the undefined-request-scoped-provider-in-a-guard gotcha in "Provider scopes", the unit/e2e split in "Testing")

Kept, on purpose:
- the sibling routing list, retitled `## Not this — route instead` — it delegates by name with a reason, which is substance, not a description echo
- all four decision tables (custom-provider form, provider scope, cross-cutting primitive, binding scope) — the flow really branches on each, and they do not overlap
- the anti-patterns table verbatim, as the rubric requires
- every code sample and the NestJS 11 / Node 20+ / Express v5 / SWC / Vitest claims

**Verification:** `bash scripts/eval-lint.sh` → `nestjs PASS (should_trigger=5, should_not_trigger=4, capability=1)`. `scripts/verify.sh` is executable and exits 0 on an empty target. Both `references/` files (`cross-cutting.md`, `testing-recipes.md`) remain linked inline from the body. `manifest.json` untouched.


ericrisco:

PR #196, opened 2026-07-29
refactor(ruby): trim description and body to the Claude-5 rubric

ericrisco:
Context-engineering pass on `ruby` against `scripts/skill-rubric.md`. Format only — no
recommendations, versions, commands, or figures changed.

| | Before | After |
|---|---|---|
| description | 605 chars | 343 chars |
| SKILL.md | 14033 bytes / 316 lines | 12763 bytes / 295 lines |

**Removed**

- The `Triggers: '…'` list of 10 phrases in three languages. It was in context on every turn
  the skill was installed, and the model routes by meaning. The `Use when` lead plus the
  `NOT Rails/ActiveRecord/ActiveSupport (that is \`rails\`)` boundary now carry discrimination.
- The `## When to use / When NOT to use` section — its Use-when bullets restated the
  description. The delegation half survives as one prose line in the intro, and every sibling
  named there is now a resolving link: `rails`, `python`, `go`, `rust`, `elixir`, `php`,
  `typescript`, `github-actions`.
- The trailing `## References` index: both reference files were already linked inline at point
  of use. Its one orphan pointer (RuboCop vs Standard config) moved into the linting section,
  so no information was dropped.

**Kept deliberately**

- The anti-patterns table — required by dimension 7, and its rows name concrete failure modes
  (`method_missing` without `respond_to_missing?`, bare `rescue Exception`, uncommitted
  `Gemfile.lock` for an app), not rules restated from above.
- The four decision tables where the flow genuinely branches: proc vs lambda, Enumerable by
  intent, `Data` vs `Struct`, Minitest vs RSpec.
- The bad/good Enumerable pair (teaches judgment by demonstration) and the Ruby 4.0 / 3.4
  hygiene notes.
- No result-envelope block added — this skill never had one.

**Verify**

- `scripts/eval-lint.sh` → `ruby PASS (should_trigger=7, should_not_trigger=4, capability=1)`
- Both `references/` files linked from the body; `scripts/verify.sh` already executable and
  makes no assertion about the description.

### DIFF
--- a/skills/ruby/SKILL.md
+++ b/skills/ruby/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: ruby
-description: "Use when writing idiomatic Ruby outside Rails — plain scripts, CLIs, gems, libraries — or refactoring imperative loops into Enumerable chains, designing module mixins, packaging with Bundler, deciding whether metaprogramming is worth it, or setting up Minitest/RSpec. Triggers: 'write this in Ruby', 'refactor this loop with map/reduce', 'build a gem', 'gemspec / Gemfile.lock', 'method_missing or define_method?', 'Minitest vs RSpec', 'FrozenError after Ruby 3.4 upgrade', 'mixin vs inheritance', 'escribir una gema en Ruby', 'refactoritzar amb blocs'. NOT Rails/ActiveRecord/controllers (that is rails)."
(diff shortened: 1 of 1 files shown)


ericrisco:

PR #197, opened 2026-07-29
refactor(procurement): Claude-5-era context pass

ericrisco:
## What

Context-engineering pass on `skills/procurement` per `scripts/skill-migration-brief.md`. No content or recommendation changed.

| | Before | After |
|---|---|---|
| `description` | 953 chars | **400 chars** |
| `SKILL.md` | 13,777 bytes / 179 lines | **12,842 bytes / 174 lines** |

## What went

- The `Triggers:` list — eight quoted phrases across English, Spanish and Catalan. Pure per-turn cost: the model routes by meaning, and every phrase restated the clause immediately before it.
- The trailing `## references/` index. Both reference files are already linked inline where they are used (`references/sourcing-requests.md` at the sourcing-request section; `references/scorecard-and-tco.md` at the scorecard, negotiation and cadence sections), so the invariant holds without the tail.

## What stayed, and why

- Kraljic 2x2 segmentation, the RFI vs RFQ vs RFP decision table, and the re-source trigger list — the flow genuinely branches there.
- Both worked bad/good pairs (unit price vs year-1 TCO; the weighted scorecard where the pricier supplier wins) — they teach the judgment by demonstration.
- The 2/10-net-30 annualized-return derivation, the maverick-spend figures (~1.8% of purchase value, ~16% of negotiated savings, <10% target) and the single/sole/dual-source distinction — load-bearing numbers.
- The anti-patterns table, unchanged: it was already `Anti-pattern | Why it bites | Do instead`, with no borrowed-urgency or excuse framing to strip.
- The hand-off routing block to `contracts`, `pricing`, `inventory`, `logistics-ops`, `invoicing`, `cost-tracking` — all verified present under `skills/`.
- No result-envelope block existed; none was invented.

## Also

Reworded two `evals/cases.yaml` `why` fields that called their cases "Spanish trigger" / "Catalan trigger" — they pointed at a trigger list that no longer exists. The prompts and expectations are untouched. `manifest.json` not touched.


ericrisco:

PR #198, opened 2026-07-29
refactor(together-fireworks): Claude-5-era description + drop duplicate reference index

ericrisco:
## Numbers

| | Before | After |
|---|---|---|
| `description` chars | 773 | 346 |
| SKILL.md bytes | 14076 | 13404 |
| SKILL.md lines | 181 | 176 |

eval-lint: **PASS** (should_trigger=7, should_not_trigger=6, capability=1)

## What went

- The `Triggers:` list — nine quoted phrases across English, Spanish and Catalan. It sat in context on every turn the skill was installed, invoked or not, and the model routes by meaning anyway. The kept description is what it does, when to use it, and the two boundaries that discriminate: `runpod` (renting GPUs to self-host) and `ollama` (local, zero marginal cost). Both verified to exist under `skills/`.
- The trailing `## References` index. Both files were already linked inline where they are needed — `references/models-and-pricing.md` under "Pick the model", `references/batch-and-tuning.md` in the Together-batch gotcha. The only detail the tail carried that the inline mention did not (dedicated-deployment notes) was folded into the inline sentence, so every `references/` file is still linked from the body.

## What stayed, and why

- **Anti-patterns table** — every row names a concrete failure mode (bare model id 404s, OpenAI Batch client against Together, budgeting at Priority rates), not a rule restated from above. The rubric requires one.
- **Both endpoint tables** and the cross-provider id mapping — the `base_url` / namespace shapes are the whole point of the skill.
- **Serving-mode decision table** — the flow genuinely branches (serverless / batch / dedicated).
- **Dated pricing rows** with their `(projected)` tags, the "read the provider page" instruction, and the Fireworks size-tier fallback — dimension 4 material, left byte-for-byte.
- **Worked cost arithmetic** and the bad → good pairs, which teach by demonstration.

No recommendations, versions, commands, figures or claims changed — this is a context pass, not a content rewrite. `scripts/verify.sh` was already executable and makes no assertion about the description. `manifest.json` untouched.


ericrisco:

PR #199, opened 2026-07-29
refactor(level-design): migrate to the Claude-5-era skill rubric

ericrisco:
## Numbers

| | Before | After |
|---|---|---|
| `description` | 972 chars | 370 chars |
| `SKILL.md` | 18,646 bytes / 175 lines | 15,073 bytes / 143 lines |

eval-lint: **PASS** (`should_trigger=7, should_not_trigger=6, capability=3`).

## What went

- The `Triggers:` phrase list (8 quoted phrases) — the model matches by meaning, and the body already taught semantic routing next to it.
- `When to use / When NOT to use` — restated the description and re-explained the same three boundaries that `Related skills` already owns.
- `The craft laws (non-negotiable)` — a rule bank restating the flow. The unique wording moved next to the step it governs (metrics-before-geometry into the pipeline, "know where to go, never how it ends" into spatial composition, "every claim is a hypothesis" into playtesting, teach-through-space into encounters, rise-and-fall into pacing).
- `Quick reference` — nine rows restating the section defaults plus links to the three references, all still linked inline at point of use.
- The trailing `References:` index line in `Related skills`.

## What stayed, and why

- **All 9 anti-pattern rows**, re-columned to `Anti-pattern | Do instead` with the `rationalizations -> STOP` framing dropped. The rows are concrete failure modes, not restated rules.
- The **pipeline stage table** — the flow genuinely branches on each stage gate.
- The four-beat encounter progression, arena rules, guidance/readability catalog, telemetry reading, reward & secret placement, the 11-item checklist, and the `02-DOCS` grounding block — all verbatim.

## Boundary

The description now names only real siblings: `game-design` (mechanics/economy), `game-storytelling` (story beats), `godot`/`unity`/`unreal` (engine tooling). All three `references/` files remain linked from the body. `manifest.json` untouched. No substance, version, or recommendation changed.


ericrisco:

PR #202, opened 2026-07-29
refactor(youtube-api): Claude-5-era description + body pass

ericrisco:
## Numbers

| | before | after |
|---|---|---|
| description chars | 1015 | 350 |
| SKILL.md bytes | 18533 | 15871 |
| SKILL.md lines | 291 | 272 |
| eval-lint | PASS | PASS (7 / 6 / 1) |

## What went

- The `Triggers:` list — seven phrases across EN/ES/CA in the description, in context on every turn the skill is installed, invoked or not.
- The `## When to use / When NOT` bullet list, which restated the description.
- The `One line: …` summary under the routing table (the boundary, stated a third time).
- The trailing `## Cross-references` index — every link already appears inline at point of use. `knowledge-ops`, the only id it added, became a routing-table row.
- Quota math stated three times with the same citation. §3 keeps the operational rule and points at §7; §7 keeps the full quote and sources.

## What stayed, and why

- **Anti-patterns table** — the rubric requires one, and these rows are concrete failure modes (committed `token.json`, partial `videos.update`, overwritten wiki snapshot), not rules restated from above.
- **Symptom → cause → fix error map** — a debugging lookup, complementary to the anti-patterns table rather than a duplicate of it.
- Scope table, quota/allocation table, the bad/good pairs for scopes and `videos.update`, and the wiki entry template: all branch points or teaching-by-demonstration.
- Routing table kept in full and upgraded from bare `(catalog id)` annotations to real `../<id>/SKILL.md` links; all ten targets verified to exist.

No recommendation, version, figure, command or source changed — the 2026-06-01 quota citation and the ~1,600-unit history are intact. `manifest.json` untouched. All three `references/` files remain linked inline (`oauth-setup.md` §1, `analytics-queries.md` §5, `wiki-schema.md` §6). `scripts/verify.sh` was already executable and carries no `Triggers:` assertion.


ericrisco:

PR #203, opened 2026-07-29
refactor(spring-boot): Claude-5-era context pass

ericrisco:
## What

Context-engineering pass on `spring-boot` against `scripts/skill-rubric.md`. No content rewrite: no recommendation, version, command, figure or claim changed.

| Metric | Before | After |
|---|---|---|
| `description` chars | 875 | 345 |
| SKILL.md bytes | 18212 | 16153 |
| SKILL.md lines | 376 | 350 |
| eval-lint | PASS | PASS (7 / 5 / 1) |

## Removed

- The 11-phrase `Triggers:` list in the description. It is in context on every turn the skill is installed, invoked or not, and the model matches by meaning. What remains: what it does, when to use it, and the `java` / `postgresdb` boundaries that actually discriminate it.
- The `When to use` section — a restatement of the description surface for surface.
- The 8-item `Decision rules` bank. Each rule was already stated inline where it applies; the three not represented in the anti-patterns table (thin controller, typed `@ConfigurationProperties`, `@Valid` at the edge) became rows there, and the rest moved into the prose of the section they govern.
- The trailing `See also` index, folded into `Boundaries` (renamed from `When NOT to use`) so every sibling link survives at point of use.

## Kept deliberately

- The anti-patterns table — retitled from `Anti-patterns / rationalizations -> STOP`, dropping the borrowed urgency framing while preserving all 12 rows plus the 3 merged in. It names concrete failure modes, which is what the rubric requires.
- The slice-selection decision table (the flow genuinely branches), the Quick reference idiom lookup, all four bad/good pairs, the pinned-stack block with its Boot 4.0 GA date, and every code sample.
- The three inline `references/` links (`jpa.md`, `security.md`, `testing.md`) — all still linked, so none is dead weight — and the project-grounding note.

## Fixes along the way

- Bare `django` backtick is now a real `../django/SKILL.md` link like its siblings.
- Two `evals/cases.yaml` `why` fields referenced "a named decision rule"; reworded to point at the sections that survive. Cases themselves unchanged.

`scripts/verify.sh` was already executable and does not assert anything about `Triggers:`. `manifest.json` untouched.


ericrisco:

PR #226, opened 2026-08-17
feat(gates): pay the gate-honesty debt — GAP state, gradeable rules, four-input refuters

ericrisco:
Closes the four things [#225](https://github.com/ericrisco/rsc-harness/pull/225) (v1.0.12) declared as debt and did not pay. A declared debt that isn't paid in the next delivery stops being a debt and becomes a note nobody reads.

**One measurement corrects the previous spec.** #225 spoke of "the `verify.sh` of four stacks" and implied a widespread fail-open. Measured: there are **189** `verify.sh` in the catalog, all distinct, and only **2** actually run coverage (`fastapi`, `go`). The problem was real and considerably smaller and sharper than that prose suggested.

## A — a coverage layer now declares whether it could have failed

`fastapi/scripts/verify.sh` ran `pytest --cov` and delegated the threshold to the project. Reasonable — until the project sets none, at which point pytest prints a percentage, exits 0, and the step records `PASS`. `go/scripts/verify.sh` is blunter: `go test -cover` takes no threshold flag, so coverage could **never** fail it. Both are the canonical shape old-coder warned about, living inside our own generator, and both fail in the direction no failing run can reveal.

Each now reports a fourth state, `[gap]`: **the layer ran and had nothing to fail with**, with its reason and a fix. The success line narrows from "all checks passed" to "every check that could fail, passed", because with a gap present the stronger claim is false in exactly the way this change exists to stop.

**A GAP deliberately does not change the exit code.** These scripts are generated for third-party repos that pass today; making them fail would publish a breaking change without warning (P7). So that it does not end up decorative (P2), the blocking lives where the user contract is: `verify` must treat a GAP as an **unverified criterion**, which already fails the verdict. *The script reports; the skill blocks.*

## B — GAP maps onto `SUSTITUIDA`, and a fourth label is banned

A layer that ran without being able to fail is not `NO-APLICA` (the surface exists) nor `HERRAMIENTA-AUSENTE` (the tool ran). It is `SUSTITUIDA`: what ran was **a report in place of a gate**, and what it cannot detect is **coverage regressing**. Stated explicitly, with a ban on inventing a fourth state — a three-label taxonomy grows on contact otherwise.

Plus the argument for changed-line coverage where it belongs: a global floor barely moves when your change lands untested (200 uncovered lines in a 20k-line tree is a rounding error against 85%), so `diff-cover --fail-under=100` gates what actually matters. **Guidance, explicitly not wiring** — the generated scripts do not do this, and the test enforces that we say so.

## C — review's refuters get four inputs, and record blind first

They received *"only the diff"*. That is not a detail: with only the diff, the one failure class no test suite can structurally catch — *the change is correct and is not what the approved spec asked for* — is invisible to the entire panel.

Now four enumerated inputs: the **task contract** (original request **plus every scope change a human explicitly approved since** — without those, a legitimate revision reads as a spec gap and you get a confident false positive), the **approved spec**, the **exact source state**, the **entry point**. With an explicit withheld list (our conversation, reasoning, defences, draft verdict) and the reason: if a claim needs our justification to stand, it is not proven.

**Blind first, compare second** — record what you attacked and found *before* seeing our conclusions; append-only after. Skip it and the fresh context is spent confirming our framing, which is the one thing it was bought to avoid. And the attack list is a deliverable: "nothing found" without saying where you looked is indistinguishable from not looking.

## D — implement forbids manufacturing green

The four moves that turn red into green without fixing anything: weakening a test, editing test and implementation in the same step, mocking the unit under test, chasing the coverage number. Each with its reason inline. No anti-pattern rows added: `author-skill/SKILL.md:125` forbids the rationalization bank that restates rules already in the flow, and it is paid for on every load.

## E — the 1.0.12 rules become gradeable, not just present

13 behavioural `must_include` items across five `cases.yaml`, written as behaviour a grader can mark present or absent ("does not present coverage as proof the suite detects bugs…", "asks for approval AGAIN…") rather than as keyword mentions. The existing eval harness can now measure whether the rules change behaviour.

## Verification

| Gate | Result |
|---|---|
| `npm run validate` | frontmatter OK |
| `npm run manifest:check` | manifest OK (264 skills) |
| `npm test` | **370 passed, 0 failed** (+19) |
| `bash scripts/eval-lint.sh` | 264 skills, 0 failures |
| `npm run drift:check` | all claims resolve |
| `bash -n` both scripts | OK |
| manual mutation | **17/17 killed** (after fixing 4 vacuous assertions) |
| P5 ceiling | max 387 (`implement`), all under 400 |

**The user-facing invariant was proven live**, not reasoned about — the real script against toy projects: gap appears with no threshold, absent with `--cov-fail-under=90`, **exit 0** when everything that ran passed beside a gap, **exit 1** when a real test broke beside one. So a GAP neither breaks a passing repo nor masks a failure. Commands are in the verification record.

**The control found four vacuous assertions, three of them the same defect as last time**: the assertion matched a *different occurrence* of the phrase than the one carrying the rule — `four inputs` appeared twice in `review`, `mutation` twice in a capability block, and `could fail, passed` matched the *comment* explaining the wording instead of the success line. The fourth was my own fix for the third, anchored to the wrong `printf`. All re-anchored to structure. Twice now is a pattern, not an anecdote, and it is written into the pattern registry: **in a prose test, anchor to structure — never to a phrase that can appear twice.**

**Declared limits, stated plainly.** The behavioural eval is **written and not run** — 13 items make the rules measurable; running it costs model calls and is the user's cost decision. *Measurable is not measured.* And `go/scripts/verify.sh` **has never been executed**: Go is not installed on this machine, so only its syntax and text are verified. Nothing ran in its place. No independent adversarial review (session policy on subagents); self-review plus 17 mutants stood in.

Three files in the working tree — `README.md`, `site/og.png`, `site/og.svg` — are **not part of this delivery** and are deliberately left uncommitted rather than swept in.

### DIFF
--- a/skills/fastapi/scripts/verify.sh
+++ b/skills/fastapi/scripts/verify.sh
@@ -9,9 +9,20 @@
 # Auto-detects each tool. If a tool is missing it prints a yellow SKIP and continues
 # (it never FAILs on a missing tool). Prefers `uv run <tool>` when `uv` is present,
 # else the bare tool on PATH. Exits non-zero only if a tool actually ran and reported
-# a failure. The coverage threshold is read from the project's pyproject.toml
-# (--cov-fail-under); this script does not hardcode a second threshold. Idempotent:
-# re-running yields the same result (read-only beyond whatever the project's pytest does).
+# a failure. Idempotent: re-running yields the same result (read-only beyond whatever
+# the project's pytest does).
+#
+# COVERAGE IS ONLY A GATE IF YOUR PROJECT MAKES IT ONE. The threshold is deliberately
+# delegated to the project (--cov-fail-under in pyproject.toml / setup.cfg / .coveragerc)
+# rather than hardcoded here. But when the project sets none, `pytest --cov` prints a
+# percentage and exits 0 however far coverage falls — and a layer that prints a number and
+# exits 0 is a report, not a gate. That case is reported as [gap] instead of PASS, because
+# the two are indistinguishable on screen and the broken one can only ever produce green:
+# no failing run will ever surface it.
+#
+# A [gap] does NOT change the exit code — breaking every repo that passes today is not this
+# script's call. This script reports; the `verify` skill judges, and there a gap counts as
+# an unverified criterion, which fails the verdict.
 #
 # Compatible with stock macOS bash 3.2: no `mapfile`, no associative arrays, and every
 # array access is guarded so `set -u` never trips on an "unbound" empty array.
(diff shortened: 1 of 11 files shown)


ericrisco:
**Correction to this PR's description.**

The body says the behavioural eval "costs model calls and is the user's cost decision". That framing was wrong, and so was the command recorded alongside it in the verification record.

Read from the code rather than inferred:

- `scripts/skill-behavior-eval.js` **invokes no model**. It is a pure scoring CLI — reads raw JSON from a file or stdin, calls `scoreFromRaw`/`holdoutGate`, exits 0/1/2. Its usage is `--score|--holdout <raw.json|->`.
- The model actually runs from `scripts/skill-behavior-eval.workflow.js`, which is a **`Workflow` tool script** (`export const meta`, `agent()`, `pipeline()`, `parallel()`). Grepping both files for `ANTHROPIC_API_KEY|api_key|anthropic|fetch|http|sdk` returns **zero matches**. There is no external API and no separate bill: it runs inside Claude Code as subagents, against the session's own usage.
- Real cost, counted: 3 agents per scenario (treatment + baseline in parallel, plus the blind grader) plus 1 loader per run. For the five skills touched here: 6 scenarios → **23 agents across 5 workflow runs**.
- And the actual blocker was never cost: this session carries an explicit instruction not to invoke `Agent` or workflows without the user asking.

Worth noting *how* this surfaced: not by any gauntlet layer, but by the user asking whether it really cost anything. No layer here executes the commands our prose cites, so a record can name a command that does not exist and stay green — the same class old-coder's late rounds kept finding, where SPEC and EVIDENCE describe something the code does not do. A check that every command cited in a record is real and invocable does not exist yet; `drift-check` resolves links, not commands. Logged as a candidate mechanism.

The spec, the plan's sibling record, and `decisions.md` have been corrected.

PR #232, opened 2026-08-18
feat(capabilities): read the ledger nobody was interrogating

ericrisco:
Spec 2 of 3. The always-on rule already asks *"did you just deliver a repeatable procedure?"* — a judgment about **one** episode, made by the agent that just performed it, at the moment it is most convinced its work was special. Grepping the always-on layer for `repeat|repetit|twice|recurr` returns **nothing**: nothing looks backwards.

The registry that would allow it **already exists**. `capabilities gap-log` writes one line per observed procedure; the only thing consuming it is `countGaps()`, which returns a **total** for `doctor` to print. The ledger is written and nobody interrogates it — the passive form of the same P2 defect.

## The evidence is the session that wrote the spec

Four manual mutation runs (8, 17, 12, 6 mutants) and five identical score-then-rescore passes. The rule was loaded in context on every turn and **never fired**, because each episode looked like one-off work on its own. Two of those four runs used a badly-escaped grep and measured wrong — errors an automated procedure would not have had.

## What lands

**Grouping calibrated, not guessed.** The first version used Jaccard at 0.6 and grouped **nothing** — the real case came out as four groups of one. So three metrics were measured on real descriptions (3 same-procedure pairs, 12 different):

| metric | same | different | margin |
|---|---|---|---|
| jaccard | 0.17–0.30 | ≤0.09 | 0.08 |
| overlap | 0.29–0.50 | ≤0.20 | 0.09 |
| **dice** | **0.29–0.46** | **≤0.17** | **0.12** ← widest |

Dice at **0.25**, with numerals and filler dropped first (*"eight mutants"* and *"twelve mutants"* are the same work, and dropping them is what made the margin measurable). The band 0.17–0.25 is **uncertain: it does not group, and records what it refused to assume** — grouping wrongly fires an offer about something that never repeated, the more expensive of the two mistakes (P7).

**Honest limit, in the code:** n=15 pairs from one session. Calibration, not validation. The measurement table is the record of how the number was chosen.

**Global count with a per-repo breakdown** (clarify): a procedure repeated across three projects is *stronger* evidence, and the per-repo log made that invisible by construction. Nothing "wins" when the two logs disagree — they are **summed**, de-duplicated by date+procedure+verdict, so double-writing cannot double-count. A home directory that cannot be written never takes down the local record.

**A sixth verdict: `proposed-capability`.** None of the original five could express *"this needs an executable capability, not a skill"* — the majority of what the detector actually finds. Without it the honest case had to be recorded as one of the five, and all five would have been false. Additive: the original five keep validating.

**`recommendKind`, with the criterion written**: capability (execute-and-compare) → agent (look without inheriting the asker's context) → skill (guidance on work already in flight). Capability is checked **first** because it is the majority case and writing a paragraph where a script belongs is the failure the generalization gate already named.

**Silence is the default.** Below threshold: nothing. Already covered: nothing (answered by reading the disk, never from memory). Previously declined: **nothing, forever** — a "no" that must be repeated weekly is what makes someone switch the harness off.

## Verification

| Gate | Result |
|---|---|
| `validate` / `manifest:check` | OK, 264 skills |
| `npm test` | **438 passed, 0 failed** (+20) |
| `eval-lint.sh` | 264 skills, 0 failures |
| `drift:check` | all claims resolve |
| manual mutation | **12/12 killed** |

Every rule tested in **both directions** (v1.0.17), and for a detector whose whole risk is nagging, the silence half is the important half. Mutant **M3 restores the Jaccard-at-0.6 that grouped nothing** — it dies against the real descriptions, which is the test that would have caught the original mistake.

End-to-end: three sightings across two repos reach the threshold and offer a **capability**, naming which repos.

## Note on how this was built

Another session is committing in the same working tree — its commits landed on top of mine on my own branch, and `README.md`/`site/*` were mid-edit. Rather than risk sweeping its work into mine, this was built in an **isolated git worktree**. Its tree was never touched.


ericrisco:
**Two fixes after the first CI failure, both mine, both worth naming.**

**1. A committed symlink broke CI.** This was built in an isolated git worktree (another session is committing in the main tree), and that worktree symlinks `02-DOCS` — which is gitignored (P9). The symlink got committed; in CI it resolves to nothing, so `node --test` walked the tree and died with `Could not find .../02-DOCS`. Untracked.

**2. The branch would have REVERTED the other session's work.** It was cut from `origin/main` at `d84acaa`; main has since gained three commits — two README/schematic changes from the other session plus release 1.0.21. So `origin/main..HEAD` showed their newer work as *removals*, and squash-merging would have undone it. Merged main in; the diff is now **only my five files**, and 438 tests are green after the merge.

Worth stating plainly: isolating in a worktree avoided clobbering their **uncommitted** work, and did nothing about their **committed** work — a branch cut from a moving main still needs to catch up before it merges. Isolation is not the same as being current.

